PR #30, opened 2019-06-03
Advanced error modeling

ctoomey:
PR resulting from [this thread](https://groups.google.com/d/topic/grpc-io/p_gCk1bn2JE/discussion).

### DIFF
--- a/content/docs/guides/error.md
+++ b/content/docs/guides/error.md
@@ -26,10 +26,9 @@ the gRPC data format (whether protocol buffers or something else). You
 may have noticed that it's quite limited and doesn't include the
 ability to communicate error details.
 
-If you're using protocol buffers as your data format, however, many of
-the [gRPC language libraries](https://grpc.io/docs/reference/) now
-provide experimental support for the richer error model developed and
-used by Google as described
+If you're using protocol buffers as your data format, however, you may
+wish to consider using the the richer error model developed and used
+by Google as described
 [here](https://cloud.google.com/apis/design/errors#error_model). This
 model enables servers to return and clients to consume additional
 error details expressed as one or more protobuf messages. It further
@@ -40,9 +39,13 @@ violations, and stack traces). The protobuf binary encoding of this
 extra error information is provided as trailing metadata in the
 response.
 
-**TODO: this [SO post](https://stackoverflow.com/a/48750825) states
-that "Support is coming to each language to handle the type". What can
-be said here about that commitment and status/timeline?**
+This richer error model is already supported in the C++, Go, Java,
+Python, and Ruby libraries, and at least the grpc-web and Node.js
+libraries have open issues requesting it. Other language libraries may
+add support in the future if there's demand, so check their github
+repos if interested. Note however that the grpc-core library written
+in C will not likely ever support it since it is purposely data format
+agnostic.
 
 You could use a similar approach (put error details in trailing
 response metadata) if you're not using protocol buffers, but you'd
--- a/content/docs/guides/error.md
+++ b/content/docs/guides/error.md
@@ -52,19 +52,16 @@ in order to make practical use of it in your APIs.
 There are important considerations to be aware of when deciding whether to
 use such an extended error model, however, including:
 
-* library implementations of the extended error model may not be consistent
+  * library implementations of the extended error model may not be consistent
 across languages in terms of requirements for and expectations of the error
 details payload
-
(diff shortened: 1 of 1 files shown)


srini100:
@ctoomey, thanks for following up with this PR. I found this [gRFC on Python](https://github.com/grpc/proposal/blob/master/L44-python-rich-status.md) implementation to be very useful to understand the extended error model. Perhaps you could work in some details from this gRFC into this PR?

content/docs/guides/error.md
```diff
@@ -18,6 +18,37 @@ If an error occurs, gRPC returns one of its error status codes instead, with an
...
+is supported by all gRPC client/server libraries, and is independent of
+the gRPC data format (whether protocol buffers or something else). You
+may have noticed that it's quite limited and doesn't include the
+ability to communicate error details.
+
+If you're using protocol buffers as your data format, however, many of
+the [gRPC language libraries](https://grpc.io/docs/reference/) now
+support the richer error model developed and used by Google as
+described
+[here](https://cloud.google.com/apis/design/errors#error_model). *TODO:
```
carl-mastrangelo: @srini100  can comment on this.

content/docs/guides/error.md
```diff
@@ -18,6 +18,37 @@ If an error occurs, gRPC returns one of its error status codes instead, with an
...
+for accessing this data in order to make practical use of it in your
+APIs.
+
+One thing to be aware of if you adopt such an extended error model,
+however, is that existing proxies, loggers, and other standard HTTP
+request processors don't have visibility into the error details and
+thus wouldn't be able to leverage them for monitoring or other
+purposes.
+
+*TODO What other caveats, if any, should be mentioned here?*
```
carl-mastrangelo: Big one:  Additional error detail in the trailers interferes with head-of-line blocking, and will cause large cache misses in the HPACK table.  Also, larger error bodies sometimes run into protocol limits (like max headers size), effectively losing the original error.  We have seen teams putting 8K of errors in each response.

ctoomey:
@srini100  I added a bit more information from the python gRFC, but given the brevity of the error page it's on I didn't want to add too much. Not sure what specifically you were thinking I'd add though so let me know.

content/docs/guides/error.md
ctoomey: @carl-mastrangelo Added this, thanks.

content/docs/guides/error.md
```diff
@@ -18,6 +18,52 @@ If an error occurs, gRPC returns one of its error status codes instead, with an
...
+types](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto)
+to cover the most common needs (such as invalid parameters, quota
+violations, and stack traces). The protobuf binary encoding of this
+extra error information is provided as trailing metadata in the
+response.
+
+**TODO: this [SO post](https://stackoverflow.com/a/48750825) states
+that "Support is coming to each language to handle the type". What can
+be said here about that commitment and status/timeline?**
+
```
srini100: @ctoomey, since there is no timeline to add this in all languages (regardless of what the comment says in the post), it won't be easy to get an answer to this question. Please either remove this TODO or replace it with something like "Currently, there is no plan to natively support this in all languages as some gRPC implementations do not take a dependency on protocol buffers. Some relevant information can be found [here](https://github.com/grpc/proposal/blob/master/L44-python-rich-status.md) and  [here](https://jbrandhorst.com/post/grpc-errors/)."

ctoomey: @srini100 from digging around in the various github repos it seems this is already supported in C++, Go, Java, Python, and Ruby, and there are open issues requesting and discussing it for at least Node and grpc-web. I'm guessing it's only the grpc-core library in C that doesn't depending on protobuf, so support might still be added for the remaining languages?

It'd be a lot more helpful to developers to provide this additional detail. How 'bout this?

> This richer error model is already supported in the C++, Go, Java, Python, and Ruby libraries, and at least the grpc-web and Node.js libraries have open issues requesting it. Other language libraries may add support in the future if there's demand, so check their github repos if interested. Note however that the grpc-core library written in C will not likely ever support it since it is purposely data format agnostic.

srini100: This looks fine.

ctoomey:
Thanks for reviewing and merging this @srini100 and @carl-mastrangelo .

PR #40, opened 2019-07-18
Typographic changes in the Supported Platforms table

dh--:
Removes a trailing | - looks like this was leftover from a copy/paste job
Replaces the * callout next to PHP with the intended meaning; alternatives would be to include the callout later on the page.
Added an empty table cell to the last row for consistency
Updated some whitespace to keep things internally consistent

### DIFF
--- a/content/about.md
+++ b/content/about.md
@@ -123,13 +123,12 @@ Check out what people are saying below:
 </tr>
 
 <tr>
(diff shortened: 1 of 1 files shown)


ejona86:
@stanley-cheung, I was thinking PHP isn't Beta. Is the asterisk because it is client-only?

ejona86:
It looks like the asterisk _was_ because it was beta. But I didn't think it was still beta.

ejona86:
@dh--, thank you!

PR #47, opened 2019-08-22
 Use proto message Get methods in Go quick start

le-michael:
Suggest users to use the exported Get methods instead of directly accessing the data.

Submitted a pull request to update the example as well.
https://github.com/grpc/grpc-go/pull/2990


menghanl:
@srini100 This PR is good to go.

PR #74, opened 2019-12-08
Fix typo in Ruby auth example.

catkins:
The auth guide has a Ruby example that references a non-existent `GRPC::Core::Credentials` class.

The class to use instead is `GRPC::Core::ChannelCredentials` as defined here:
https://github.com/grpc/grpc/blob/bebd20b126303a1f5d6cbb31d0463460b59c5074/src/ruby/ext/grpc/rb_channel_credentials.c#L237


apolcyn:
this is ok to merge, I don't have  merge ability though

catkins:
Thanks @apolcyn! Who should I tag to get it merged?

PR #78, opened 2020-01-31
Add get_grpcurl installer

gnossen:
This PR introduces an experimental one-line installer for `grpcurl`. It has been manually tested on Mac OS and several linux distros. For the moment, this is completely unsupported and undocumented.

The intent is to enable people to run `curl -s https://grpc.io/get_grpcurl | bash`. For obvious reasons, this script does not require root.

CC @lidizheng

### DIFF
--- a/static/get_grpcurl
+++ b/static/get_grpcurl
@@ -3,7 +3,7 @@
 set -e
 
 UNATTENDED="false"
-if [ "$#" = "1" -a "$1" = "-y" ]; then
+if [ "$#" = "1" ] && [ "$1" = "-y" ]; then
   UNATTENDED="true"
 elif [ "$#" -gt 1 ]; then
   echo "Usage: $0 [-y]" 1>&2
@@ -43,7 +43,8 @@ confirm () {
     return 0
   fi
   echo "$1"
-  read -p "[y/n]" -r RESPONSE </dev/tty
+  printf "[y/n]"
+  read -r RESPONSE </dev/tty
   printf "\n"
   case "$RESPONSE" in
     "yes" | "y" | "YES" | "Y" ) return 0;;
@@ -65,7 +66,7 @@ modify_config_file () {
   FILE="$1"
   BIN_DIR="$2"
   BACKUP="${FILE}.bak.$(date | sed 's/ /-/g')"
-  if ! stat ${FILE} 1>/dev/null 2>&1; then
+  if ! stat "${FILE}" 1>/dev/null 2>&1; then
     return 1
   fi
   cp "${FILE}" "${BACKUP}"
--- a/static/get_grpcurl
+++ b/static/get_grpcurl
@@ -1,0 +1,77 @@
+#!/bin/sh
+
+set -e
+
+detect_os () {
+  RAW_OS=$(uname -s)
+  OS="linux"
+  case "$RAW_OS" in
+    "Linux") OS="linux";;
+    "Darwin") OS="osx";;
+    *) echo "Unrecognized operating system \"${RAW_OS}\"" 1>&2; exit 1;;
+  esac
+  echo "${OS}"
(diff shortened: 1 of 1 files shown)


static/get_grpcurl
```diff
@@ -0,0 +1,77 @@
+#!/bin/sh
```
lidizheng: LICENSE?

static/get_grpcurl
```diff
@@ -0,0 +1,77 @@
...
+  echo "${OS}_${ARCH}"
+}
+
+confirm () {
+  echo "$1"
+  read -p "[y/n]" RESPONSE </dev/tty
+  printf "\n"
+  case "$RESPONSE" in
+    "yes" | "y" ) return 0;;
+    "no" | "n" ) echo "Aborting." 1>&2; exit 1;;
```
lidizheng: How about uppercase?

static/get_grpcurl
gnossen: The other files in this repo don't appear to have a license file.

gnossen:
@lidizheng Shell check is a nice tool. I learned a couple of things from it.

static/get_grpcurl
```diff
@@ -0,0 +1,77 @@
...
+}
+
+GITHUB_URL="https://github.com"
+GRPCURL_REPO="fullstorydev/grpcurl"
+
+PLATFORM=$(detect_platform)
+
+BIN_DIR="${HOME}/.grpcurl/bin"
+
+confirm "This script will perform a user install of grpcurl. Would you like to proceed?"
```
nicolasnoble: I would add a command line argument to skip the confirmation, to allow unattended installation.

static/get_grpcurl
```diff
@@ -0,0 +1,77 @@
...
+mkdir -p "${BIN_DIR}"
+
+curl -Ls "${GITHUB_URL}/${GRPCURL_REPO}/releases/latest" | \
+  grep "href=.*${PLATFORM}.tar.gz" | \
+  cut -d '"' -f 2 | \
+  echo "${GITHUB_URL}$(cat -)" | \
+  curl -sL "$(cat -)" --output - | \
+  tar -C "${BIN_DIR}" -xzf - grpcurl
+
+cp "${HOME}/.bashrc" "${HOME}/.bashrc.bak.$(date | sed 's/ /-/g')"
```
nicolasnoble: This file may not always be present. You need to check for .profile too. Not everyone uses bash, and MacOS is removing it.

static/get_grpcurl
```diff
@@ -0,0 +1,77 @@
...
+GITHUB_URL="https://github.com"
+GRPCURL_REPO="fullstorydev/grpcurl"
+
+PLATFORM=$(detect_platform)
+
+BIN_DIR="${HOME}/.grpcurl/bin"
+
+confirm "This script will perform a user install of grpcurl. Would you like to proceed?"
+
+ensure_command "curl"
```
nicolasnoble: You also technically need to ensure cut, grep, tar, mkdir and cp.

gnossen: Fair enough. Shouldn't there be some sort of "GNU collection standard"? I know there's a POSIX shell standard, but it seems like there should also be some lowest common denominator that busybox et al. can conform to.

gnossen:
@nicolasnoble Thanks for the review! Good points. PTALA.

static/get_grpcurl
nicolasnoble: GNU's coreutils will usually be always installed, that's true, and has cut, mkdir and cp. Tar and grep... I've seen oddballs distros without them by default.

Might as well be thorough anyway, it's not costing anything :-)

nicolasnoble:
Thanks!

PR #80, opened 2020-02-04
Correct release tag

gnossen:
This is an alternative to https://github.com/grpc/grpc/pull/21891.

Tested locally.


srini100:
Works after merging.

Issue #132, opened 2020-03-13
Kotlin gRPC


chalin:
- [x] Add placeholder pages for the quick start and tutorial so that we can link to them from other sites before the final versions are published - #141
- [x] **About** page supported languages and platforms
  - [x] Add Kotlin entry - (#141), #142
  - [x] Remove "under development" qualifier - https://github.com/grpc/grpc.io/commit/59e517353b2e6dfe553248a09fce8c90209606d6
- [x] Create Kotlin **quick start** - (#141), #124, #194
- [x] Create Kotlin basic **tutorial** (`docs/tutorials/basic/kotlin.md`) - (#141), #157, #194
- [x] API reference:
  - [x] Ensure API reference has been generated and is accessible; see https://github.com/grpc/grpc-kotlin/issues/5.
  - [x] Ensure that there is a link to the [Kotlin API reference][] from the [Reference page](https://grpc.io/docs/reference/) - #165, #175
- [x] Add Kotlin entry to [packages.grpc.io](https://packages.grpc.io) - https://github.com/grpc/grpc/pull/22451
- [x] Add Kotlin to  [grpc/grpc README](https://github.com/grpc/grpc) - https://github.com/grpc/grpc/pull/22450
- [x] Merge `kotlin` branch into `master`

After launch:
- [x] Blog post on grpc.io referring to the [Google Cloud blog entry](https://cloud.google.com/blog/products/application-development/use-grpc-with-kotlin) (here's the [tweet](https://twitter.com/grpcio/status/1250821545036877829)) - #196
- [x] Delete [grpc-java/examples/example-kotlin](https://github.com/grpc/grpc-java/tree/master/examples/example-kotlin) - https://github.com/grpc/grpc-java/pull/6936

Questions:
- Do we want to provide some guidance in terms of writing unit tests for grpc-kotlin apps (and write an example for the Hello World app)? Cf. grpc-java/examples/example-kotlin's [Unit test examples](https://github.com/grpc/grpc-java/tree/master/examples/example-kotlin#unit-test-examples)

Answered:
- [x] On which platforms (Windows, Linux, Mac) can grpc-kotlin be used? 
- [x] Router guide example:
  - [x] Stick with proto `Duration` type, even if it isn't used in other examples?
  - [x] Drop/cleanup Java specific `option` declarations?

[Kotlin API reference]: https://grpc.github.io/grpc-kotlin/grpc-kotlin-stub

chalin:
[...]

PR #142, opened 2020-03-17
Kotlin/JVM platforms

chalin:
Contributes to #132


chalin:
@lowasser, like this?

lowasser:
Yes, that's precisely what I had in mind.

PR #172, opened 2020-03-27
C++ quick start: fix instructions

### LINKED ISSUE #95: c++ quickstart instructions refer only to make, but then point at a doc for bazel
https://grpc.io/docs/quickstart/cpp/ says "follow the instructions to install gRPC C++ via make", but the [linked page](https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make) leads with bazel (with a secondary reference to cmake).

### LINKED ISSUE #97: protobuf build/install instructions broken
https://grpc.io/docs/quickstart/cpp/ directs users to build protobufs by using make.
There are two problems with this:
1) protobufs are a submodule - I reckon not everyone knows how to clone a repo WITH the submodules
2) there's no Makefile - it looks like it's been moved over to bazel

chalin:
Closes #95
Closes #97

cc @srini100 @jtattermusch @veblush

### DIFF
--- a/content/docs/quickstart/cpp.md
+++ b/content/docs/quickstart/cpp.md
@@ -9,63 +9,111 @@ description: This guide gets you started with gRPC in C++ with a simple working 
 
 #### gRPC
 
-To install gRPC on your system, follow the [instructions to install gRPC C++ via make](https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make).
+Follow the gRPC installation instructions for your OS:
 
-To run the example code, please ensure `pkg-config` is installed on your
-machine before you build and install gRPC in the previous step, since the
-example `Makefile`s try to look up the installed gRPC path using `pkg-config`.
-On Debian-based systems like Ubuntu, this can usually be done via
-`sudo apt-get install pkg-config`.
+- macOS:
+
+  ```sh
+  $ brew install grpc
+  ```
+
+- Any OS: to build and install from sources, refer to [Building from
+  source][BUILDING].
 
 #### Protocol Buffers v3
 
-While not mandatory to use gRPC, gRPC applications usually leverage Protocol
-Buffers v3 for service definitions and data serialization, and our example code
-uses Protocol Buffers as well as gRPC. If you don't already have it installed on
-your system, you can install the version cloned alongside gRPC. First ensure
-that you are running these commands in the gRPC tree you just built in the from
-the previous step.
+While not mandatory, gRPC applications usually leverage [Protocol Buffers
+v3][pbv3] for service definitions and data serialization, and the example code
+uses Protocol Buffers.
+
+- macOS (if you installed `grpc` using brew, then you can skip this step since
+  `protobuf` is installed as a prerequisite):
+
+  ```sh
+  $ brew install protobuf
+  ```
+
+- Any OS:
+
+   1. Download a zip file of the latest version of pre-compiled binaries for
+      your operating system from [github.com/google/protobuf/releases][]
+      (`protoc-<version>-<os>.zip`).
+   2. Unzip the file.
+   3. Update your environment's path variable to include the path to the
+      `protoc` executable.
+
+If you've built gRPC from sources, you have another option. You can install
+protobuf from its submodule in the `third_party` folder:
 
 ```sh
 $ cd third_party/protobuf
 $ make && sudo make install
 ```
 
-###  Build the example
+#### pkg-config
 
-Always assuming you have gRPC properly installed, go into the example's
-directory:
+Follow the installation instructions for your OS:
+
+- Linux:
+
+  ```sh
+  $ sudo apt-get install pkg-config
+  ```
+
+- macOS:
+
+  ```sh
+  $ brew install pkg-config
+  ```
+
+### Download the example
+
+You can skip this step if you've installed gRPC from sources.
+
+The example code is available from the [grpc repo][repo]. Clone the repo at the
+latest release:
 
 ```sh
-$ cd examples/cpp/helloworld/
+$ git clone -b {{< param grpc_release_tag >}} https://github.com/grpc/grpc
 ```
 
-Let's build the example client and server:
+###  Build the example
+
+Change to the example's directory:
+
+```sh
+$ cd examples/cpp/helloworld
+```
+
+Run make to build the example client and server:
+
 ```sh
 $ make
 ```
 
-Most failures at this point are a result of a faulty installation (or having
-installed gRPC to a non-standard location. Check out [the installation
-instructions for details](https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make)).
+{{< note >}}
+  **Getting build failures?** Most issues, at this point, are a result of a
+  faulty installation, or having installed gRPC to a non-standard location.
+  Carefully recheck your installation.
+{{< /note >}}
+
 
 ### Try it!
 
-From the `examples/cpp/helloworld` directory, run the server, which will listen
-on port 50051:
+From the `examples/cpp/helloworld` directory:
 
-```sh
-$ ./greeter_server
-```
+ 1. Run the server:
 
-From a different terminal, run the client:
+    ```sh
+    $ ./greeter_server
+    ```
 
-```sh
-$ ./greeter_client
-```
+ 1. From a different terminal, run the client and see the client output:
 
-If things go smoothly, you will see the `Greeter received: Hello world` in the
-client side output.
+    ```sh
+    $ ./greeter_client
+    Greeter received: Hello world
+    ```
 
 Congratulations! You've just run a client-server application with gRPC.
 
@@ -74,11 +122,11 @@ Congratulations! You've just run a client-server application with gRPC.
 Now let's look at how to update the application with an extra method on the
 server for the client to call. Our gRPC service is defined using protocol
 buffers; you can find out lots more about how to define a service in a `.proto`
-file in [What is gRPC?](/docs/guides/) and [gRPC Basics:
-C++](/docs/tutorials/basic/cpp/). For now all you need to know is that both the server and the client
-"stub" have a `SayHello` RPC method that takes a `HelloRequest` parameter from
-the client and returns a `HelloResponse` from the server, and that this method
-is defined like this:
+file in [What is gRPC?](/docs/guides) and [gRPC Basics:
+C++](/docs/tutorials/basic/cpp). For now all you need to know is that both the
+server and the client stub have a `SayHello()` RPC method that takes a
+`HelloRequest` parameter from the client and returns a `HelloResponse` from the
+server, and that this method is defined like this:
 
 ```protobuf
 // The greeting service definition.
@@ -100,8 +148,8 @@ message HelloReply {
 
 Let's update this so that the `Greeter` service has two methods. Edit
 `examples/protos/helloworld.proto` (from the root of the cloned repository) and
-update it with a new `SayHelloAgain` method, with the same request and response
-types:
+update it with a new `SayHelloAgain()` method, with the same request and
+response types:
 
 ```protobuf
 // The greeting service definition.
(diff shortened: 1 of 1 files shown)


chalin:
Related: https://github.com/grpc/grpc/issues/22419

content/docs/quickstart/cpp.md
```diff
@@ -9,63 +9,111 @@ description: This guide gets you started with gRPC in C++ with a simple working
...
 
-To install gRPC on your system, follow the [instructions to install gRPC C++ via make](https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make).
+Follow the gRPC installation instructions for your OS:
 
-To run the example code, please ensure `pkg-config` is installed on your
-machine before you build and install gRPC in the previous step, since the
-example `Makefile`s try to look up the installed gRPC path using `pkg-config`.
-On Debian-based systems like Ubuntu, this can usually be done via
-`sudo apt-get install pkg-config`.
+- macOS:
```
jtattermusch: nit: It feels a bit odd to include macOS as a special case.

content/docs/quickstart/cpp.md
```diff
@@ -9,63 +9,111 @@ description: This guide gets you started with gRPC in C++ with a simple working
...
+
+  ```sh
+  $ brew install grpc
+  ```
+
+- Any OS: to build and install from sources, refer to [Building from
```
jtattermusch: A better entry point for installing gRPC C++ from the users's perspective (as opposed to being a gRPC contributor) is actually https://github.com/grpc/grpc/tree/master/src/cpp#to-start-using-grpc-c (which then points to building.md for more details).

content/docs/quickstart/cpp.md
```diff
@@ -9,63 +9,111 @@ description: This guide gets you started with gRPC in C++ with a simple working
...
+   2. Unzip the file.
+   3. Update your environment's path variable to include the path to the
+      `protoc` executable.
+
+If you've built gRPC from sources, you have another option. You can install
+protobuf from its submodule in the `third_party` folder:
 
 ```sh
 $ cd third_party/protobuf
 $ make && sudo make install
```
jtattermusch: I wouldn't recommend to install  protobuf library  at a system level (=i.e. without a prefix & using a local directory). There is no ways of uninstalling and it will forever pollute your system with a specific version of protobuf (which will cause a lot of problems once trying to upgrade protobuf)

content/docs/quickstart/cpp.md
```diff
@@ -9,63 +9,111 @@ description: This guide gets you started with gRPC in C++ with a simple working
...
+  $ brew install protobuf
+  ```
+
+- Any OS:
+
+   1. Download a zip file of the latest version of pre-compiled binaries for
+      your operating system from [github.com/google/protobuf/releases][]
+      (`protoc-<version>-<os>.zip`).
+   2. Unzip the file.
+   3. Update your environment's path variable to include the path to the
```
jtattermusch: while getting the protoc binaries that way is straighforward and useful, it won't really help you build a simple grpc application because you'll also need libprotobuf, libprotoc and the gRPC protoc plugins to be able to build your example app.
And once you're able to get libprotobuf built and installed, you can trivially build and install the protoc binaries.

content/docs/quickstart/cpp.md
jtattermusch: Ideally, one would install both gRPC and protobuf using one of the approaches mentioned in
https://github.com/grpc/grpc/tree/master/src/cpp#cmake

When installing, one should always use CMAKE_INSTALL_PREFIX to avoid a system-wide installation.

jtattermusch: Btw, with newer version of cmake, once can easily install both protobuf and grpc by a single command.
https://github.com/grpc/grpc/tree/master/src/cpp#fetchcontent

content/docs/quickstart/cpp.md
chalin: Agreed, but I wouldn't want to omit the brew-based install for macOS users merely because it is the _only_ platform that supports installation other than from sources.

Are you aware of a similar way to install binaries under Linux and/or Windows?

content/docs/quickstart/cpp.md
chalin: Done

content/docs/quickstart/cpp.md
chalin: > I wouldn't recommend to install protobuf library at a system level ...

Ok, I've removed these instructions for now.

> Ideally, one would install both gRPC and protobuf using one of the approaches mentioned in ... 
> ... [or using fetchcontent]

Agreed, but I'd like to propose that we address that in the next round of edits. WDYT @jtattermusch?

content/docs/quickstart/cpp.md
chalin: I've tested this approach under macOS and it worked fine for me, as far as I recall. I agree that this page will require further edits, but I seen this as I first round of changes. Can we keep this text as is for now?

If you prefer that I remove these lines I can. I'd just like to get the remaining edits is so that we can start another pass.

content/docs/quickstart/cpp.md
chalin: Ok, I've updated the instructions to explain how to do a local install using `CMAKE_INSTALL_PREFIX`. PTAL

content/docs/quickstart/cpp.md
chalin: Removed

content/docs/quickstart/cpp.md
chalin: Remove

chalin:
I've rewritten the setup instructions. They now walk the user through a build and local install of gRPC and protobuf. PTAL

chalin:
This is still WIP, but I wanted to get your initial feedback on the direction that it is heading in. (What is missing is listing the gRPC build prerequisite tools: build-essential autoconf libtool pkg-config and libssl-dev.)

chalin:
Page updates are complete. PTAL

content/docs/quickstart/cpp.md
```diff
@@ -3,69 +3,163 @@ layout: quickstart
...
+$ wget -q -O cmake-linux.sh https://github.com/Kitware/CMake/releases/download/v{{< param cmake-version >}}/cmake-{{< param cmake-version >}}-Linux-x86_64.sh
+$ sh cmake-linux.sh -- --skip-license --prefix=$MY_INSTALL_DIR
+$ rm cmake-linux.sh
 ```
 
-Most failures at this point are a result of a faulty installation (or having
-installed gRPC to a non-standard location. Check out [the installation
-instructions for details](https://github.com/grpc/grpc/blob/master/src/cpp/README.md#make)).
+
+#### gRPC and Protocol Buffers v3
```
jtattermusch, line 71: note that protobuf 3.x.y (which is what some people will understand under protocol buffers v3) and "proto3" are not the same thing.   First is the version of protobuf release and  "proto3" is the new syntax for .proto files.

Given that proto3 support has been around for several years now, I think it's fine to omit the "v3" from the section title. (maybe add a note that the recommended syntax is proto3).

content/docs/quickstart/cpp.md
```diff
@@ -3,69 +3,163 @@ layout: quickstart
...
+          -DgRPC_BUILD_TESTS=OFF \
+          -DCMAKE_INSTALL_PREFIX=$MY_INSTALL_DIR \
+          ../..
+    $ make -j
+    $ make install
+    $ popd
+    ```
+
+You can find a complete set of instructions for building gRPC C++ in [Building
+from source][from-source].
```
jtattermusch, line 115: note that https://github.com/grpc/grpc/blob/master/src/cpp/README.md provides the instructions for building from the user's perspective (and BUILDING.md is mostly meant for power users and contributors, as noted in its first paragraph).

content/docs/quickstart/cpp.md
```diff
@@ -127,32 +219,34 @@ Remember to save the file!
 
 ### Generate gRPC code
```
jtattermusch, line 220: nit: "Regenerate gRPC code" may be better because we're generating again after we've updated the .proto file?

content/docs/quickstart/cpp.md
```diff
@@ -3,69 +3,163 @@ layout: quickstart
...
+    ```sh
+    $ cd examples/cpp/helloworld
+    ```
+
+ 2. Build the example using `cmake`:
+
+    ```sh
+    $ mkdir -p cmake/build
+    $ pushd cmake/build
+    $ cmake ../..
```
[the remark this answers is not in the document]
chalin: My understanding is that `CMAKE_INSTALL_PREFIX` is to be used when we want to install something from a package, but we aren't installing anything from the example.

I've retested the setup instructions and they work as given, provided the `PATH` variable is properly set. Recall that one of the steps earlier in the page is:

```console
$ export PATH="$PATH:$MY_INSTALL_DIR/bin"
```

If I omit this step, then running `cmake` on the example fails:

```console
cmake ../..
-- The C compiler identification is GNU 7.5.0
...
CMake Error at CMakeLists.txt:103 (find_package):
  Could not find a package configuration file provided by "Protobuf" with any
  of the following names:

    ProtobufConfig.cmake
    protobuf-config.cmake

  Add the installation prefix of "Protobuf" to CMAKE_PREFIX_PATH or set
  "Protobuf_DIR" to a directory containing one of the above files.  If
  "Protobuf" provides a separate development package or SDK, be sure it has
  been installed.
-- Configuring incomplete, errors occurred!
...
```

After I set the `PATH` variable, then `cmake ../..` finds `protoc` and `grpc`:

```console
cmake ../..
CMake Warning (dev) at /home/chalin/local/lib/cmake/protobuf/protobuf-options.cmake:6 (option):
  Policy CMP0077 is not set: option() honors normal variables.  ...
-- Using protobuf
-- Using gRPC 1.28.1
-- Configuring done
-- Generating done
-- Build files have been written to: /home/chalin/grpc/examples/cpp/helloworld/cmake/build
```

content/docs/quickstart/cpp.md
chalin: Done

content/docs/quickstart/cpp.md
chalin: The README provides instructions on how to _use_ gRPC. In the quick start we are instructing readers to locally _build_ gRPC, so I'd rather keep the link to BUILDING page. WDYT?

content/docs/quickstart/cpp.md
chalin: Done

content/docs/quickstart/cpp.md
jtattermusch: sg, I'll leave up to you. Just wanted to point out the dichotomy.

jtattermusch: On the other hand, we should at least provide a link to README.md because it describes other modes of installation with cmake (fetchcontent, as submodule, etc).

content/docs/quickstart/cpp.md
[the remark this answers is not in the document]
jtattermusch: I feel that setting PATH variable is dirty and it's not the right cmake way of doing it.

Actually one of the error messages you posted gives a hint about what to do (*Add the installation prefix of "Protobuf" to CMAKE_PREFIX_PATH or set
  "Protobuf_DIR" to a directory containing one of the above files*).

In one of our distribtests we're actually just setting CMAKE_INSTALL_PREFIX and it works just fine
https://github.com/grpc/grpc/blob/80b74811d466db2f7cfed06f0f3b33395b15c876/test/distrib/cpp/run_distrib_test_cmake.bat#L86

If everything works after setting CMAKE_PREFIX_PATH, I'd use that because it seems to be the cleanest option.
https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html

content/docs/quickstart/cpp.md
chalin: Done

content/docs/quickstart/cpp.md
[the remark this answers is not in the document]
chalin: Yes it works with `CMAKE_PREFIX_PATH`. I've updated the command to make use of it.

chalin:
@jtattermusch - done, PTAL

jtattermusch:
@chalin  can you merge yourself or do I need to do it?

chalin:
I can merge, thanks.

PR #173, opened 2020-03-27
Dart quick start: update protobuf installation instructions

chalin:


chalin:
Although we talk about templatizing the quick start and tutorial, we might just want to have a separate Package Installation section. (After this PR lands :))

chalin:
Merging these rather small edits. I can address review comments later, if necessary.

PR #183, opened 2020-04-10
Reworked Go quick start

chalin:
Might help alleviate #177?

@srini100 @jtattermusch @ejona86 - who best should be assigned to review this?

cc @lucperkins @zacharysarah @thisisnotapril


content/docs/quickstart/go.md
```diff
@@ -7,9 +7,8 @@ description: This guide gets you started with gRPC in Go with a simple working e
 
 ### Prerequisites
 
-- Go version 1.6 or higher.
-
-For installation instructions, see Go's [Getting Started](https://golang.org/doc/install) guide.
+- **Go** version 1.9+. For installation instructions, see Go's [Getting
```
chalin, line 10: Could we just require v1.11+ and use modules throughout this page?

ejona86:
@dfawley, reassign as appropriate.

content/docs/quickstart/go.md
dfawley: Yes.  1.9 is no longer officially supported (although we still run tests to make sure it works).  Using modules would require some changes to these instructions, but that would be appreciated.  Also, I'd recommend documenting to set `GO111MODULE=on` for consistency regardless of the version of Go.

chalin: Great, I can update the text now before you review further if you'd like. Let me know what you prefer.

chalin: Actually, what should be the minimum version of Go that we document as a prerequisite?

chalin: Note that https://github.com/grpc/grpc-go/blob/master/README.md contains:

> **Prerequisites**
gRPC-Go requires Go 1.9 or later.

That should be updated too if 1.9 is no longer officially supported.

dfawley: We support the latest 3 versions of Go, so right now: 1.12, 1.13, and 1.14.

Thanks for pointing out the `README.md`.  I'll update it.

dfawley: > Great, I can update the text now before you review further if you'd like. Let me know what you prefer.

I would prefer for you to update this before review, since I think some of the changes will be pretty substantial.  Thank you!

chalin: I'm committed WIP for this change.

content/docs/quickstart/go.md
```diff
@@ -141,26 +142,33 @@ message HelloReply {
...
-Next we need to update the gRPC code used by our application to use the new
-service definition. From the same examples dir as above
-(`$GOPATH/src/google.golang.org/grpc/examples/helloworld`):
+Before you can use the new service method, you need to recompile the updated
+proto file.
+
+From the `examples/helloworld` directory, run the following command:
 
 ```sh
 $ protoc -I helloworld/ helloworld/helloworld.proto --go_out=plugins=grpc:helloworld
```
chalin, line 155: @dfawley - It seems that source files fetched as modules are read only. I'm getting an error because the `.pb.go` file cannot be overwritten:

```console
$ protoc -I helloworld/ helloworld/helloworld.proto --go_out=plugins=grpc:helloworld
helloworld/helloworld.pb.go: Permission denied
```

Should I be fetching the example source in a different way?

content/docs/quickstart/go.md
dfawley: (FYI https://github.com/grpc/grpc-go/pull/3516)

content/docs/quickstart/go.md
dfawley: I think it would be best to copy the helloworld example to a new directory as part of these instructions (just the `.proto` and `.go` files?).

chalin: @dfawley - Note that the client and server have imports like this:

```go
import ( ...
  pb "google.golang.org/grpc/examples/helloworld/helloworld"
)
```

So, if I copy the files to a local directory and create the corresponding module and try to run the server I get an "ambiguous import" error:

```console
$ cd "$HOME/examples/helloworld"
$ go mod init google.golang.org/grpc/examples/helloworld
$ go run greeter_server/main.go
greeter_server/main.go:30:2: ambiguous import: found package google.golang.org/grpc/examples/helloworld/helloworld in multiple modules:
        google.golang.org/grpc/examples/helloworld (/home/chalin/examples/hello_world/helloworld)
        google.golang.org/grpc v1.28.1 (/home/chalin/go/pkg/mod/google.golang.org/grpc@v1.28.1/examples/helloworld/helloworld)
```

Given https://github.com/golang/go/issues/32128, I believe that we don't currently have a way to resolve this ambiguity, right?

chalin: I'm experimenting with creating a local module (for the copied sources) and changing the import paths. Does that sound like a sound approach? It seems to be working so far.

chalin: @dfawley - done, PTAL

content/docs/quickstart/go.md
```diff
@@ -3,98 +3,150 @@ title: Go Quick Start
...
+    ```sh
+    $ go mod init examples/helloworld
+    ```
+
+ 6. Adjust import paths to reference local packages (rather than those from the
+    original `google.golang.org/grpc` module):
+
+    ```sh
+    $ perl -pi -e 's|google.golang.org/grpc/||g' greeter_{client,server}/main.go
+    ```
```
chalin, line 119: @dfawley - it would be great if the original example sources were already setup in this way (so that we can avoid these last two steps). Might that be possible?

chalin:
@dfawley - ETA on a review?

chalin:
@dfawley - btw, we have a "gRPC Docs Working Group" GVC at 13:30 ET in case you'd like to join. Or, if you prefer to have a focused discussion concerning this PR with me, I'm available for a VC. Otherwise, comments attached to this PR would be fine. Let me know what would be simplest for you.

dfawley:
> @dfawley - btw, we have a "gRPC Docs Working Group" GVC at 13:30 ET in case you'd like to join. Or, if you prefer to have a focused discussion concerning this PR with me, I'm available for a VC. Otherwise, comments attached to this PR would be fine. Let me know what would be simplest for you.

I had already started to review this, so I finished up & sent my comments.  I can join the GVC if you'd like to discuss anything significant, but I didn't have any major comments, so I think the PR should be sufficient.  Thanks!

zacharysarah:
@chalin 
> we have a "gRPC Docs Working Group" GVC at 13:30 ET in case you'd like to join.

I’d love to join! Is there an invitation?

content/docs/quickstart/go.md
```diff
@@ -3,98 +3,150 @@ title: Go Quick Start
...
 description: This guide gets you started with gRPC in Go with a simple working example.
+protoc-version: 3.11.4
 ---
 
 ### Prerequisites
 
-- Go version 1.6 or higher.
-
-For installation instructions, see Go's [Getting Started](https://golang.org/doc/install) guide.
+- **Go** version 1.12+. For installation instructions, see Go's [Getting
```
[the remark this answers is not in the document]
chalin: Done

content/docs/quickstart/go.md
```diff
@@ -3,98 +3,150 @@ title: Go Quick Start
...
 
 #### gRPC
 
-Use the following command to install gRPC.
+Use the following command to install gRPC as a [module][]:
 
 ```sh
-$ go get -u google.golang.org/grpc
+$ export GO111MODULE=on # Enable module-aware mode
+$ go get -u google.golang.org/grpc@{{< param grpc_release_tag >}}
```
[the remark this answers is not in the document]
chalin: Ok (I was having issues earlier -- it would pick up the package at head? -- but that was before I switched to using modules).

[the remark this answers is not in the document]
chalin: Done

chalin:
@zacharysarah - I've asked @jtattermusch to add you to the weekly meeting invite.

content/docs/quickstart/go.md
```diff
@@ -3,98 +3,150 @@ title: Go Quick Start
...
-`.proto` file using the protocol compiler: `protoc`.
+ 2. The `protoc` plugin for Go (`protoc-gen-go`) was installed as a dependency
+    of the `grpc` module. You can confirm this, or install the plugin, using the
+    following command:
 
-For the purpose of this example, the `helloworld.pb.go` file has already been
-generated (by compiling `helloworld.proto`), and can be found in this directory:
-`$GOPATH/src/google.golang.org/grpc/examples/helloworld/helloworld`
+    ```sh
+    $ go get -u github.com/golang/protobuf/protoc-gen-go
```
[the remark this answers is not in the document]
chalin: It seems to happen as part of the `get`:

```console
$ go get github.com/golang/protobuf/protoc-gen-go
go: downloading github.com/golang/protobuf v1.4.0
go: found github.com/golang/protobuf/protoc-gen-go in github.com/golang/protobuf v1.4.0
go: downloading google.golang.org/protobuf v1.21.0
$ ls $(go env GOPATH)/bin
protoc-gen-go
```

content/docs/quickstart/go.md
dfawley: I missed this comment somehow.  So make the import paths relative?  I think that would be OK if it's not considered an anti-pattern (we always use absolute paths).

content/docs/quickstart/go.md
```diff
@@ -3,98 +3,150 @@ title: Go Quick Start
...
+While not mandatory, gRPC applications usually leverage [Protocol Buffers][pb]
+for service definitions and data serialization, and the example code uses
+[proto3][].
 
-  * Unzip this file.
-  * Update the environment variable `PATH` to include the path to the protoc binary file.
+ 1. Install the `protoc` compiler:
 
-Next, install the protoc plugin for Go
+    - macOS:
```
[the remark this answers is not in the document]
chalin: Ok, I've reworked this part to mention the use of Homebrew at the end.

content/docs/quickstart/go.md
```diff
@@ -3,98 +3,150 @@ title: Go Quick Start
...
+          (`protoc-<version>-<os><arch>.zip`). For example:
 
-### Download the example
+          ```sh
+          $ PB_REL="https://github.com/protocolbuffers/protobuf/releases"
+          $ curl -LO $PB_REL/download/v{{< param protoc-version >}}/protoc-{{< param protoc-version >}}-linux-x86_64.zip
+          ```
 
-The grpc code that was fetched with `go get google.golang.org/grpc` also contains the examples. They can be found under the examples dir: `$GOPATH/src/google.golang.org/grpc/examples`.
+       2. Unzip the file under `$HOME/local` or a directory of your choice. For
```
chalin, line 48: @dfawley: I've changed `$HOME/local` to `$HOME/.local` (notice the leading dot). Any objections?

content/docs/quickstart/go.md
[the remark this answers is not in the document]
chalin: Actually, I'll revert this to using the tag so that it will be easier to specify the path to the example source below. Also, some of the other quick start and tutorial pages refer to a specific release. (Release numbers are updated often enough :))

chalin:
@dfawley - done with changes PTAL

content/docs/quickstart/go.md
chalin: My concern is that having the reader follow 6 steps _just_ to copy an example's source seems like a lot of work -- especially for a _quick_ start :). Is there any way that we can avoid some of these 6 steps?

My original suggestion what that the `helloworld` example be defined as a separate module already (named, say, `examples/helloworld`) in the `grpc-go` repo, but I don't know if it makes sense to do so (that is, to define a module inside of a module).

content/docs/quickstart/go.md
dfawley: Sounds fine - I don't have any strong feelings on this.

content/docs/quickstart/go.md
dfawley: To be fair, one whole step is changing directories.

Is there a curl command we can use to fetch the examples from github?  That would avoid the need to chmod them.

The "Download" link on any directory in github links you to something like this: https://download-directory.github.io/?url=https://github.com/grpc/grpc-go/tree/master/examples/helloworld

But then it asks you for a token. :unamused:

> My original suggestion what that the helloworld example be defined as a separate module already (named, say, examples/helloworld) in the grpc-go repo, but I don't know if it makes sense to do so (that is, to define a module inside of a module).

Yes, it's totally OK to have modules inside other modules.  Such a change is a breaking change, though.  But this is just an example.  And there's some amount of shenanigans you may need to do to avoid the conflict of older versions of grpc-go having the same packages.  I think you do this by setting the sub-module to require a version of grpc-go _after_ (or at the same commit where) the sub-module was forked.  But I'm not able to find the right documentation right now (I'll keep searching).

dfawley: Of course, after 5 minutes of failed searches....then my first query after sending this turned up what I was looking for:

https://github.com/golang/go/wiki/Modules#is-it-possible-to-add-a-module-to-a-multi-module-repository

chalin: Thanks for the link.

If you are in general agreement with the changes made in this PR, I'd like to get it merged before moving on to a possibly simpler solution. Would you agree?

dfawley: I'm generally fine with the contents of this PR.  One thing I dislike is the use of `perl`, here, which is not installed by default on all systems.  Can this be changed to use `sed` instead?

chalin: On other projects I've encountered problems with the use use `sed -i` -- e.g., see this [SO post](https://stackoverflow.com/questions/9704020/recursive-search-and-replace-in-text-files-on-mac-and-linux). That's why I'm proposing the use of perl here.

(I've just tried it again and you must use `-i ''` under macOS and `-i` (w/o any argument) under Linux.

chalin: (Hopefully the use of this command will only be temporary :)).

chalin:
@lucperkins - did you want to review this?

chalin:
Thanks for the approval @dfawley!

content/docs/quickstart/go.md
chalin: Here's the PR for making `helloworld` a module: https://github.com/grpc/grpc-go/pull/3540

dfawley: If step 5 was changed to:

`go mod init google.golang.org/examples/helloworld`

Then would step 6 be unnecessary?

chalin: Yes since you're altering the path to drop the "/grpc/" part. Without step 6 the code would refer to the `.pb.go` files from the official `grpc` module.

dfawley: "Yes" step 6 would be _unnecessary_?  Or "Yes" it would still be necessary?

If the former, we can just make that change and not do anything in the `grpc-go` repo (I think?).

chalin: Oops, apologies for being unclear in my answer. Step 6 would still be necessary because `go mod init google.golang.org/examples/helloworld` essentially changes the path to the example. The quick start also (now) changes the path to the example.

Btw, did you mean to write: `go mod init google.golang.org/grpc/examples/helloworld` -- which would preserve the path to the example, merely making `helloworld` an independent module? In that case we run into the ambiguous import problem.

Might you want us to discuss this (well, https://github.com/grpc/grpc-go/pull/3540, actually) over a VC some time tomorrow?

dfawley: > Btw, did you mean to write: go mod init google.golang.org/grpc/examples/helloworld -- which would preserve the path to the example, merely making helloworld an independent module? In that case we run into the ambiguous import problem.

Yes, that's what I meant, sorry.  OK, I see we still would also need a `go.mod` inside the grpc repo so that it removes those paths from its module.

At this point, given all the constraints I can see, and the options you laid out in https://github.com/grpc/grpc-go/pull/3540, what I'd prefer is adding a single `go.mod` to `grpc/examples` (3), and having the quick start guide recommend copying all of `grpc/examples` elsewhere to work on.  If that SGTY then I don't think there is any need to meet.  If you would prefer another direction or just want to explore all/some of the options further, then I'm happy to meet and am available tomorrow.  Feel free to email me (or use hangouts) and we can set up a time.

chalin: I answered in https://github.com/grpc/grpc-go/pull/3540#issuecomment-614973544 -- so that we could move the discussion back over there.

PR #184, opened 2020-04-13
Simplify syntax in example code

waldyrious:
Use a semicolon instead of an empty options list.

This is consistent with the style guide and examples and in the protobuf documentation:
- https://developers.google.com/protocol-buffers/docs/style#services
- https://developers.google.com/protocol-buffers/docs/proto3#services
- https://developers.google.com/protocol-buffers/docs/reference/proto3-spec#service_definition

### DIFF
--- a/content/docs/guides/_index.md
+++ b/content/docs/guides/_index.md
@@ -65,7 +65,7 @@ protocol buffer messages:
 // The greeter service definition.
 service Greeter {
   // Sends a greeting
-  rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc SayHello (HelloRequest) returns (HelloReply);
 }
 
 // The request message containing the user's name.
--- a/content/docs/quickstart/android.md
+++ b/content/docs/quickstart/android.md
@@ -71,7 +71,7 @@ server, and that this method is defined like this:
 // The greeting service definition.
 service Greeter {
   // Sends a greeting
-  rpc SayHello (HelloRequest) returns (HelloReply) {}
+  rpc SayHello (HelloRequest) returns (HelloReply);
 }
 
 // The request message containing the user's name.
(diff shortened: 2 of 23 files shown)


waldyrious:
For the record, the discrepancies led to some doubts about what syntax would be valid / recommended. From a brief search it looks like [others](https://stackoverflow.com/q/43771925/266309) also had the [same](https://stackoverflow.com/q/30106667/266309) doubts.

chalin:
To be fair, there isn't an explicit mention of this in the style guide. But thanks for all the links, including to SO.

I'd rather not update old blog entries. For the rest, I'd rather that the site reflect the actual sources that they are originating from. So this would mean making changes to each of the corresponding repos, and then only making the change here once the original sources have been updated.

I'd suggest that we take it one language at a time. Let me first try with the new Kotlin example, since I've done some cleanup in that `.proto` file already. I'll report back on results here.

chalin:
FYI, cleanup done for gRPC-Kotlin examples - https://github.com/grpc/grpc-kotlin/pull/62.

waldyrious:
Regarding the blog post — makes sense, I'll revert the changes to the blog post.

As for the source repositories, point taken, and thanks for the Kotlin example. Are you planning to make the changes yourself, or should I go ahead and submit PRs?

chalin:
> Are you planning to make the remaining changes yourself, ...

I'd like to propose simplifications to the `.proto` files for the two other languages I'm currently working on. Let's see how those are accepted first before tackling all other languages. Thanks.

chalin:
Hi @waldyrious: it'll take a while before this is addressed (in part because we need to resolve issues like https://github.com/grpc/grpc-proto/pull/81 first). Can you create an issue with the opening PR comment, and then close this PR? If you'd rather that I do it, let me know.

waldyrious:
Thanks for the update. I understand that coordinating changes across several repos is not straightforward.

> Can you create an issue with the opening PR comment, and then close this PR? If you'd rather that I do it, let me know.

I'd be more comfortable with you doing that, yeah — your context is more complete than mine.

chalin:
Closing. Will track progress via #282.

PR #214, opened 2020-04-23
Add a protoc installation page

chalin:
The quick start and tutorial pages duplicate `protoc` installation instructions, and not all instructions are up-to-date. This PR creates a separate installation page that we'll more conveniently be able to link to.

cc @zacharysarah

### DIFF
--- a/content/docs/protoc-installation.md
+++ b/content/docs/protoc-installation.md
@@ -1,0 +1,75 @@
+---
+title: Protocol Buffer Compiler Installation
+short: Protoc Installation
+description: How to install the protocol buffer compiler.
+protoc-version: 3.11.4
+---
+
+While not mandatory, gRPC applications often leverage [Protocol Buffers][pb] for
+service definitions and data serialization. Most of the example code from this
+site uses [version 3 of the protocol buffer language (proto3)][proto3].
+
+The protocol buffer compiler, `protoc`, is used to compile `.proto` files, which
+contain service and message definitions. Choose one of the methods given below
+to install `protoc`.
+
+### Install using a package manager
+
+You can install the protocol compiler, `protoc`, with a package manager under
+Linux or macOS using the following commands.
+
+- Linux:
+
+  ```sh
+  $ apt install -y protobuf-compiler
+  $ protoc --version  # Ensure compiler version is 3+
+  ```
+
+- MacOS, using [Homebrew][]:
+
+  ```sh
+  $ brew install protobuf
+  $ protoc --version  # Ensure compiler version is 3+
+  ```
+
+### Install pre-compiled binaries (any OS)
+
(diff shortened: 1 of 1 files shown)


chalin:
Thanks for the review @ejona86. I'm going to merge this now since I have another PR build atop this one. I'll address any suggested changes to the new `protoc` install page in a followup PR(s).

content/docs/protoc-installation.md
```diff
@@ -0,0 +1,75 @@
...
+site uses [version 3 of the protocol buffer language (proto3)][proto3].
+
+The protocol buffer compiler, `protoc`, is used to compile `.proto` files, which
+contain service and message definitions. Choose one of the methods given below
+to install `protoc`.
+
+### Install using a package manager
+
+You can install the protocol compiler, `protoc`, with a package manager under
+Linux or macOS using the following commands.
```
jtattermusch, line 19: Some notes on the linux installation:
we should ad a note that  while installing the linux package `protobuf-compiler` is generally the easiest way, it won't give you the latest released version, so you might not get access to the latest fixes and features (which might or might not be a problem, depending on your use).

Note that to be able to use protoc with grpc, besides `protoc` you'll also need the grpc-specific plugins that generate the grpc service stubs/clients. I'm not sure how useful it is to just obtain `protoc` (there seems to be protobuf-compiler-grpc package, but I'm afraid it's too old).

Current package versions in major distros:
Current debian stable is at 3.6 (which is OKish, but not great)
https://packages.debian.org/buster/protobuf-compiler
Current ubuntu stable is at 3.0.0 which is borderline (several years old).
https://packages.ubuntu.com/bionic/protobuf-compiler

A few more notes:
- The apt command only exists on debian-based distros.
- on mac brew  will give you a much newer version than apt / apt-get 
- The note "Ensure compiler version is 3+" makes sense, because anything before 3.0.0 is definitely too old
- in some languages, we provide a "tools" package that has a pre-compiled version of protoc and the corresponding grpc plugin bundled (e.g. grpcio.tools for python or Grpc.Tools for C#).

chalin:
> we should add a note that  while installing the linux package `protobuf-compiler` is generally the easiest way, it won't give you the latest released version, so you might not get access to the latest fixes and features (which might or might not be a problem, depending on your use).

I've created an issue for that: #229

> Note that to be able to use protoc with grpc, besides `protoc` you'll also need the grpc-specific plugins that generate the grpc service stubs/clients.

Each language page explains how to get the language-specific plugins.

> in some languages, we provide a "tools" package that has a pre-compiled version of protoc and the corresponding grpc plugin bundled (e.g. grpcio.tools for python or Grpc.Tools for C#).

Right, so we only refer to this `protoc` installation page for those languages/platforms where this is not the case.

PR #241, opened 2020-05-27
Go quick start: adjust protoc cmd line options

### LINKED ISSUE #236: The location of pb file generated in quickstart(go) is incorrect
**Issue description**
The location of pb file generated in quickstart(go) is incorrect

https://grpc.io/docs/quickstart/go/
1. Get the example code
```
$ git clone https://github.com/grpc/grpc-go
```
2. Regenerate gRPC code
```
$ protoc -I helloworld/ helloworld/helloworld.proto --go_out=plugins=grpc:helloworld
```
3. pb file in the wrong location
```
helloworld
├── google.golang.org
│   └── grpc
│       └── examples
│           └── helloworld
│               └── helloworld
│                   └── helloworld.pb.go   <---- 
├── helloworld.pb.go                       <---- 
└── helloworld.proto
```

**Expected behavior**
pb file can be updated correctly

chalin:
Since the last update to the Go quick start, the following option was added to the example's `helloworld.proto` file:

```proto
option go_package = "google.golang.org/grpc/examples/helloworld/helloworld";
```

This PR ensures that the generated `.pb.go` file ends up in the example's local `helloworld` folder.

Closes #236

### DIFF
--- a/content/docs/languages/go/quickstart.md
+++ b/content/docs/languages/go/quickstart.md
@@ -129,7 +129,7 @@ Before you can use the new service method, you need to recompile the updated
 While still in the `examples/helloworld` directory, run the following command:
 
 ```sh
-$ protoc -I helloworld/ helloworld/helloworld.proto --go_out=plugins=grpc:helloworld
+$ protoc -I helloworld --go_out=plugins=grpc:helloworld --go_opt=paths=source_relative helloworld.proto
 ```
 
 This will regenerate the `helloworld/helloworld.pb.go` file, which contains:
--- a/content/docs/languages/go/quickstart.md
+++ b/content/docs/languages/go/quickstart.md
@@ -129,7 +129,7 @@ Before you can use the new service method, you need to recompile the updated
 While still in the `examples/helloworld` directory, run the following command:
 
 ```sh
-$ protoc -I helloworld --go_out=plugins=grpc:helloworld --go_opt=paths=source_relative helloworld.proto
+$ protoc --go_out=plugins=grpc:. --go_opt=paths=source_relative helloworld/helloworld.proto
 ```
 
 This will regenerate the `helloworld/helloworld.pb.go` file, which contains:


chalin:
cc @ejona86

content/docs/languages/go/quickstart.md
```diff
@@ -129,7 +129,7 @@ Before you can use the new service method, you need to recompile the updated
 While still in the `examples/helloworld` directory, run the following command:
 
 ```sh
-$ protoc -I helloworld/ helloworld/helloworld.proto --go_out=plugins=grpc:helloworld
+$ protoc -I helloworld --go_out=helloworld --go_opt=paths=source_relative helloworld.proto
```
dfawley, line 132: This removed the `plugins=grpc` parameter -- isn't it needed to generate services?

chalin: Oh thanks, I had ment to ask you about that. The command ran just fin w/o it, but now that I check the diff, the core of the generated file is missing (and on second thought it makes sense that `--go_out` doesn't necessarily imply that the `grpc` plugin is being used). Oops! I'll add it back.

chalin: (And I was looking at https://developers.google.com/protocol-buffers/docs/reference/go-generated, that didn't include the plugin, but reading until the end of that page, I understand why.)

chalin: Done. PTAL

content/docs/languages/go/quickstart.md
```diff
@@ -129,7 +129,7 @@ Before you can use the new service method, you need to recompile the updated
 While still in the `examples/helloworld` directory, run the following command:
 
 ```sh
-$ protoc -I helloworld/ helloworld/helloworld.proto --go_out=plugins=grpc:helloworld
+$ protoc -I helloworld --go_out=plugins=grpc:helloworld --go_opt=paths=source_relative helloworld.proto
```
dfawley, line 132: `helloworld.proto`?  This changed from `helloworld/helloworld.proto` -- is the path to the file no longer necessary because of the change to the `-I` or the `source_relative` option?

ejona86: Actually, the `-I` shouldn't be there at all, if this is running from `examples/helloworld` and the file's path is `examples/helloworld/helloworld/helloworld.proto`. The `-I` is only to bring in other protoc roots. (Basically, you just need `-I .`, but I think that is there by default.)

chalin: Ok, here's what I've tried:

- Dropping the `-I` option or using `-I .` results in 

  ```nocode
  Could not make proto path relative: helloworld.proto: No such file or directory
  ```

- Dropping the `-I` option or using `-I .`, _and_ updating the source path to `helloworld/helloworld.proto` generates the `.pb.go` file under `examples/helloworld/helloworld/helloworld/helloworld.proto` (that's 3 `helloworld` directories, which isn't what we want). I guess that makes sense under `paths=source_relative`.

The following seems to work as desired:

```console
$ protoc --go_out=plugins=grpc:. --go_opt=paths=source_relative helloworld/helloworld.proto
```

This is what the PR currently contains:

```console
$ protoc -I helloworld --go_out=plugins=grpc:helloworld --go_opt=paths=source_relative helloworld.proto
```

Which command would you prefer that I use? (I'd vote for the former.)

dfawley: > Which command would you prefer that I use? (I'd vote for the former.)

I agree, that one seems to make more sense.

chalin: Ok, I've updated the PR. PTAL

PR #245, opened 2020-05-28
Move API-reference links to language-specific pages

chalin:
- Contributes to #181
- **Preview** (selected pages):
  - https://deploy-preview-245--grpc-io.netlify.app/docs/languages
  - https://deploy-preview-245--grpc-io.netlify.app/docs/languages/cpp
    **Note** the API-reference links to gRPC C++ and gRPC core)
  - https://deploy-preview-245--grpc-io.netlify.app/docs/languages/java
- Redirect test: https://deploy-preview-245--grpc-io.netlify.app/docs/reference

cc @thisisnotapril @jtattermusch @zacharysarah @mjpitz


content/docs/languages/csharp/dotnet.md
```diff
@@ -2,11 +2,12 @@
 title: gRPC for .NET
 ---
 
-For an introduction to gRPC for .NET ([grpc-dotnet][]) and a basic tutorial, see
-the following pages:
+For an introduction to gRPC for .NET ([grpc-dotnet][]), a basic tutorial, and
+API reference documentation, see the following pages:
```
mjpitz, line 6: This is redundant. You're just repeating the items listed below.

content/docs/languages/_index.md
```diff
@@ -5,12 +5,17 @@ weight: 1
...
+- Quick start
+- Tutorials
+- API reference
+
 Select a language to get started:
 
 - [Android Java](android)
 - [C#](csharp)
-- [C++](cpp)
+- [C++](cpp), and gRPC Core
```
mjpitz, line 18: Adding this makes it a bit inconsistent with the rest.

Maybe `[C / C++](cpp)` instead?

content/docs/languages/csharp/dotnet.md
chalin: Yeah, I was feeling that too. Fixed.

content/docs/languages/_index.md
chalin: @ejona86 - WDYT?

chalin: Or I could just remove "gRPC Core" text I added here.

content/docs/languages/csharp/dotnet.md
```diff
@@ -2,11 +2,11 @@
 title: gRPC for .NET
 ---
 
-For an introduction to gRPC for .NET ([grpc-dotnet][]) and a basic tutorial, see
-the following pages:
+The following pages cover the gRPC for .NET implementation:
```
mjpitz, line 5: `for the .NET implementation`

chalin: Yeah, that text is a bit off. I've reworked it. PTAL

content/docs/languages/_index.md
ejona86: I think we can just drop (not add) the "gRPC Core" text. Almost all the documentation is C++, and we don't really advertise the C apis anywhere. They aren't so much for users as for other implementors.

ejona86:
(Maybe it would just be enough to have the "out arrow" on the reference links, the arrow saying "you are leaving this site." I realize that isn't entirely true, but it would then be clear why the links aren't on the side navbar. There's probably at least 5 different ways this could be solved.)

content/docs/languages/_index.md
chalin: 👍 Reverting that line to what we had before.

chalin:
Thanks for the feedback. Updates done. Merging.

chalin:
As for the "out arrow" icon, I thought that we had an issue open over that. If we don't I'll create one.

chalin:
Here's the issue: _External links should be decorated with an external-link icon_ #246

PR #260, opened 2020-06-02
Add note about wait_for_termination

### LINKED ISSUE #259: Python Basics Tutorial claims you need to sleep-loop where you should actually call wait_for_termination()
**Issue description**
The Python language guide states:
```
Because start() does not block you may need to sleep-loop if there is nothing else for your code to do while serving.
```
where should actually call `wait_for_termination()`. If you indeed do sleep-loop, you'll block the main thread and get weird errors, e. g. `Connection reset by peer`.

see [https://github.com/grpc/grpc.io/blob/master/content/docs/languages/python/basics.md#L310](https://github.com/grpc/grpc.io/blob/master/content/docs/languages/python/basics.md#L310)

```python
def serve():
    server = grpc.server(ThreadPoolExecutor(max_workers=1))

    inpainting_service_pb2_grpc.add_InpainterServicer_to_server(
        InpainterServicer(), 
        server)

    server.add_insecure_port("[::]:50051")
    server.start()

serve()
while True:
    time.sleep(1)
```

**Expected behavior**
```
Because `start()` does not block make sure to call `wait_for_termination()` if there is nothing else for your code to do while serving.
```

gnossen:
Resolves #259.

### DIFF
--- a/content/docs/languages/python/basics.md
+++ b/content/docs/languages/python/basics.md
@@ -305,10 +305,14 @@ def serve():
       RouteGuideServicer(), server)
   server.add_insecure_port('[::]:50051')
   server.start()
+  server.wait_for_termination()
 ```
 
-Because `start()` does not block you may need to sleep-loop if there is nothing
-else for your code to do while serving.
+The server `start()` method is non-blocking. A new thread will be instantiated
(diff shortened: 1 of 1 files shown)


gnossen:
CC @chalin

PR #264, opened 2020-06-03
Fix side menu scrolling

### LINKED ISSUE #217: Navigation sidebar cannot easily be scrolled
I was on https://grpc.io/docs/tutorials/basic/dart/ . I noticed that the navigation sidebar was cut off. So I tried to scroll to see more options, and couldn't. Instead, I have to scroll to the very end of the page for more of the sidebar to be visible.

This makes the sidebar navigation mostly useless for languages lower on the list.

image

----------------------

image

celestehorgan:
Closes #217 

This PR introduces a max-height for the side menu (determined by the viewport height) and forces the side menu overflow to scroll. It also introduces a small margin-right on the carets/arrows in the side menu, so that they don't overlap with the scrollbar.

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -214,6 +214,8 @@ $colors: mergeColorMaps(("secondary": ($secondary, $white), "twitter-blue": ($tw
 .is-sticky
   +sticky
   top: $navbar-height + 3rem
+  height: 100vh
+  overflow-y: scroll
 
 .is-constrained
   max-width: 90ch
@@ -253,6 +255,9 @@ $colors: mergeColorMaps(("secondary": ($secondary, $white), "twitter-blue": ($tw
 
     & + &
       margin-top: 1.5rem
+
+.level-right
+  margin-right: 0.75rem
 
 .blog-list
   ul


chalin:
Hi @celestehorgan! Thanks for the PR.

assets/sass/style.sass
```diff
@@ -214,6 +214,8 @@ $colors: mergeColorMaps(("secondary": ($secondary, $white), "twitter-blue": ($tw
 .is-sticky
   +sticky
   top: $navbar-height + 3rem
+  height: 80vh
+  overflow-y: scroll
```
chalin, line 218: Nice improvement. A few comments:

- Are you sure that you want to apply these changes to everything that "is-sticky" vs. only the left-side nav? (I'm not sure what else "is-sticky" is used for.)
- The nav content gets cropped
  > image

chalin:
While this improves on #217, it isn't a complete fix. All site implementations that I've worked with make use of some JS to get this right -- that is, to avoid having that gap (https://github.com/grpc/grpc.io/pull/264#discussion_r434843220). Do you want to give that a try?

celestehorgan:
@chalin using JS to adjust viewport heights is an older way of doing this supplanted by the `vh` CSS unit. CSS is easier on load and processing time, so it's best to use the CSS unit rather than JS.

I've set it to `100vh` to further address your feedback.

chalin:
> using JS to adjust viewport heights is an older way of doing

Good to know.

> I've set it to 100vh ...

Still isn't quite working for me. Steps to reproduce:

- Make your browser window as short as possible -- mine goes as low as 296px in Chrome.
- Visit https://deploy-preview-264--grpc-io.netlify.app/docs/talks/.
- Try to scroll the sidenav up.
  What happens for me is that page content scrolls up and the sidenav stays put until I've reached the end of the page content. Are you seeing that too?

I can reproduce this behavior with other pages and taller windows.

celestehorgan:
@chalin I see the behavior you're describing... but is this really a problem worth investing time into?

At a viewport height of 290px or so, it's far more likely that the user will be served the mobile version of the site than anything. 

I can investigate further fixes to this, possibly with javascript (`vh` grabs the browser window's height _including_ the browser's menu bar, which is why things get weird at smaller resolutions), but I think it's a better idea to implement a proper mobile sub-navigation, which is something I was planning to look at next :)

chalin:
> I think it's a better idea to implement a proper mobile sub-navigation, which is something I was planning to look at next

That sounds good to me. Let's track the remaining improvements via #295.

PR #279, opened 2020-06-11
gRPC-web interceptors blog: revisions 20/06/11

chalin:
Here's a first pass over the blog entry:

- I've strived to keep the intent of the original content as best I could.
- Some material was reordered to provide (what I believe to be) a better flow and a gentler introduction to the material.

Material I've (tentatively) removed:

- Dropped mention of the details of what protoc generates.
- Dropped mention of `grpc.web.UnaryResponse` since it seemed like a lot of detail not supporting the main intent of the blog.
-  While `grpc.web.Request` is still mentioned, the reader is referred to the source file if she/he wants to know what it contains.

If you find that the blog entry doesn't read well with out the material that was removed, let me know what you'd like to add back.

@Jennnnny and @stanley-cheung: please have a look and let me know what you think. (Sorry @Jennnnny, I can't mark you as an official reviewer in the Reviewers list, but you can still review this PR and add comments.)

cc @thisisnotapril 

Note that this PR is over the [blog-grpc-web branch](https://github.com/grpc/grpc.io/tree/blog-grpc-web), so we can merge it if it is good enough, and I can continue following revisions in another PR without affecting the `master` branch.


content/blog/grpc-web-interceptor.md
```diff
@@ -1,46 +1,30 @@
...
-gRPC-Web has two clients generated by
-[Protoc](https://github.com/grpc/grpc-web#client-configuration-options) for one
-service: let’s say *FooServiceClient* and *FooServicePromiseClient*. The
-response type of *FooServiceClient* is
-[ClientReadableStream](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/clientreadablestream.js)
-for BOTH unary calls and server streaming calls, so **StreamInterceptor** is
-expected to be used for intercepting *FooServiceClient* calls. The response type
-of *PromiseClient* is
-[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise),
-so we should use **UnaryInterceptor** for *PromiseClients*.
```
chalin, line 36: FYI: these are the details of the use of `protoc` that I've dropped from this revision.

content/blog/grpc-web-interceptor.md
```diff
@@ -209,6 +163,31 @@ TestStreamInterceptor.prototype.intercept = function(request, invoker) {
...
+{{< note >}}
+  Interceptors are executed in reverse order for request processing, and in order for response processing:
+
+  ```nocode
+  Request  --> interceptor3  --> interceptor2  --> interceptor1 --> Intercepted --> Server
+                                                                    request           |
+                                                                                      |
+  Intercepted <-- interceptor3 <-- interceptor2 <-- interceptor1 <--   Response  <-- Server
+  response
+  ```
```
chalin, line 188: This ASCII art seems clunky, but I wasn't sure what to replace it buy, so I've left it in for now.

Might you be open to a more textual description that explains the order processing using a functional notation? Possibly something like this:

> To illustrate this, assume that `i1()`, `i2()` and `i3()` functions representing the effects of the interceptors bound to the client.
> - If `req` is the original client request, then the server will receive the request created by the following interceptor transformations: `i1(i2(i3(req)`.
> - Similarly, if `resp` is a server response, then the client will receive the response that results from the following order of interceptor processing: `i3(i2(i1(resp)`.

content/blog/grpc-web-interceptor.md
```diff
@@ -209,6 +163,31 @@ TestStreamInterceptor.prototype.intercept = function(request, invoker) {
...
+{{< /note >}}
+
 ### Advanced example: RetryInterceptor
```
chalin, line 191: As an introduction to grpc-web interceptors, I'd drop this "advance example".

IMHO, it might be more useful be able to include links to running examples.

content/blog/grpc-web-interceptor.md
```diff
@@ -1,46 +1,30 @@
...
-[Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise),
-so we should use **UnaryInterceptor** for *PromiseClients*.
-
-The design of UnaryInterceptor looks like:
+For example, the `UnaryInterceptor` interface is declared like this:
 
 ```js
 /*
-* StreamInterceptor shares the same semantics except the return type of it is
-* ClientReadablaStream
```
chalin, line 43: I moved this comment into the main text (leaving it hear seems too early and it feels more confusing than anything else).

content/blog/grpc-web-interceptor.md
```diff
@@ -52,88 +36,58 @@ const UnaryInterceptor = function() {};
...
-- **[grpc.web.UnaryResponse](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/unaryresponse.js)**:
-  A wrapper class of response message, status and response metadata, so that we
-  could handle the response metadata(headers) via interceptors.
+Basically, an interceptor allows you to do the following:
 
-## Configure interceptors
+- Update the original gRPC request before it is passed along. For example, you
+  might inject extra information such as auth headers.
+- Manipulate the behavior of the original invoker function, such as bypassing
+  the call or updating the response before it is returned to the client.
```
chalin, line 58: I suppose that it is possible to "bypass the RPC call" but how common and/or useful is that? Maybe we can stick to mentioning response updates?

content/blog/grpc-web-interceptor.md
```diff
@@ -52,88 +36,58 @@ const UnaryInterceptor = function() {};
...
 
-Besides grpc.web.Request, we’ve introduced another new class here:
+## What can I do with an interceptor?
 
-- **[grpc.web.UnaryResponse](https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/unaryresponse.js)**:
-  A wrapper class of response message, status and response metadata, so that we
-  could handle the response metadata(headers) via interceptors.
+Basically, an interceptor allows you to do the following:
 
-## Configure interceptors
```
chalin, line 78: This section is moved to after the examples so that the reader gets to see what an interceptor looks like before seeing code that makes use of interceptor instances.

content/blog/grpc-web-interceptor.md
```diff
@@ -142,46 +96,46 @@ TestUnaryInterceptor.prototype.intercept = function(request, invoker) {
...
+`ClientReadableStream` using a `StreamInterceptor`. These are the main steps to
+follow:
 
- 1. Create a new subclass of ClientReadableStream that wraps around the original
-    stream and overrides its methods, so that we could handle its response and
-    update callbacks(listeners).
- 2. Create a new subclass of StreamInterceptor. While implementing the
-    StreamInterceptor.prototype.intercept method, return the wrapped
-    ClientReadableStream.
- 3. Apply the instantiated StreamInterceptor during the client construction.
```
chalin, line 156: I removed this point since it is covered in the section that follows

content/blog/grpc-web-interceptor.md
```diff
@@ -209,6 +163,31 @@ TestStreamInterceptor.prototype.intercept = function(request, invoker) {
 };
 ```
 
+## Binding interceptors
```
chalin, line 166: I updated the title because "Configuring" didn't seem quite right. If you'd prefer a term other than "Binding" let me know.

content/blog/grpc-web-interceptor.md
Jennnnny: I do want to add some details on the use case of StreamInterceptor and UnaryInterceptor in this blog, since it's a very confusing concept. : (
Users will (I guess) usually think StreamInterceptor is used for streaming calls and UnaryInterceptor is for unary calls. But actually StreamInterceptor can intercept both unary and streaming calls.  Which interceptor to use is only decided by the client type they are using.
I'll also add this to the code comments as well

content/blog/grpc-web-interceptor.md
Jennnnny: Sure. Or can we add an image to the blog?

content/blog/grpc-web-interceptor.md
```diff
@@ -270,6 +249,18 @@ RetryInterceptor.prototype.intercept = function(request, invoker) {
...
+[grpc-web-team@google.com][].
+
+[1.1.0]: https://github.com/grpc/grpc-web/releases/tag/1.1.0
+[gRPC languages]: /docs/languages
+[gRPC mailing list]: https://groups.google.com/forum/#!forum/grpc-io
+[grpc-web-team@google.com]: mailto:grpc-web-team@google.com
+[grpc-web]: https://github.com/grpc/grpc-web
+[grpc.web.Request]: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/request.js
+[interceptor.js]: https://github.com/grpc/grpc-web/blob/master/javascript/net/grpc/web/interceptor.js
+[issue]: https://github.com/grpc/grpc-web/issues/new
```
Jennnnny, line 266: Those links looks great! Thanks

content/blog/grpc-web-interceptor.md
Jennnnny: We haven't ported all of our examples to Github yet. The only one is here. We may add more examples to this file or under this directory in the future. 
https://github.com/grpc/grpc-web/blob/cc6cdfb428f21dd70b9e714f0dacc0ec0745d06b/net/grpc/gateway/examples/echo/commonjs-example/client.js

content/blog/grpc-web-interceptor.md
Jennnnny: I think it could be common since people may want to use cached result instead? Another very common use case of intercepting response is implementing retry.

content/blog/grpc-web-interceptor.md
chalin: Ok. How about I add a note to this effect -- i.e., that a StreamInterceptor can be used to intercept both unary and streaming calls.

(Do you have an example of a such a StreamInterceptor?)

content/blog/grpc-web-interceptor.md
chalin: Sure an image would be ok. Maybe something like this quick sketch (with all the labels added)?

image

If you prefer an image, I'd let you create it. If you're ok with the text suggested in the previous comment in this thread, let me know.

content/blog/grpc-web-interceptor.md
chalin: So, shall we drop this example, or do you prefer to keep it?

content/blog/grpc-web-interceptor.md
chalin: Thanks, I think that mentioning use of cached results would be good. I'll add that.

chalin:
@Jennnnny - is this PR a good enough increment to warrant merging it (even though some changes are needed)? Once it is merged, either or both of us can work from the branch base.

Also, I don't mind implementing the changes that are discussed in this PR, but I won't be able to do that until some time (hopefully early) next week.

content/blog/grpc-web-interceptor.md
chalin: Actually, investigating this more, while I see that the grpc-web protoc plugin generates two clients for each service, this is nonstandard AFAIK. @stanley-cheung Is this documented anywhere?

To say that a StreamInterceptor can be used to intercept both unary and streaming calls seems misleading for someone who might be familiar with gRPC but not the particularities of grpc-web.

Ok, I'll need to give this some more thought. It might be clearer to say that a StreamInterceptor can be used for any RPC with a ClientReadableStream return type (which would be either a server-streaming RPCs and unary RPCs from the generated client that only uses ClientReadableStream for return types (is there a name for this kind of generated client?).

stanley-cheung: Yes currently the grpc-web protoc plugin generate both the normal callback-based client (which returns ClientReadableStream for both unary and streaming calls), as well as the Promise-based client which wasn't very well documented. I can see how it's difficult to convey this in this blog post. To further complicate things, the Promise-based client is only promise-based for unary calls (there is no promise-based streaming API), hence the difficult naming of `UnaryInterceptor`.

I think it will be wise to emphasize that `StreamInteceptor` is for calls that return `ClientReadableStream`.

But on the other hand, perhaps should we consider to _only_ focus this blog post on `StreamInterceptor`? This way our message can be more focused and less confusing. Perhaps we can leave Promise-based client as a totally separate topic for a later time?

chalin: > I think it will be wise to emphasize that StreamInteceptor is for calls that return ClientReadableStream.

Agreed.

> ... perhaps should we consider to only focus this blog post on StreamInterceptor

That might be more effort at this point, especially if we agree to simply emphasize that: "a `StreamInteceptor` can be applied to any call with a `ClientReadableStream` return type (whether a unary or a server-streaming RPC)"

How does that sound?

stanley-cheung: Sounds good to me. Thanks!

content/blog/grpc-web-interceptor.md
Jennnnny: Yeah I'll create an image. Where should I submit it to?

Jennnnny:
Sure. Please merge it. Thank you veri=y much!

content/blog/grpc-web-interceptor.md
chalin: You can submit a PR over the `blog-grpc-web` branch. Place the image under `static/img`. Thanks!

content/blog/grpc-web-interceptor.md
Jennnnny: You can drop this example. I think we can add it somewhere under https://github.com/grpc/grpc-web/blob/cc6cdfb428f21dd70b9e714f0dacc0ec0745d06b/net/grpc/gateway/examples/echo/commonjs-example/

content/blog/grpc-web-interceptor.md
chalin: Added a note.

content/blog/grpc-web-interceptor.md
chalin: (I added a phrase about caching.)

content/blog/grpc-web-interceptor.md
chalin: Thanks for the image.

content/blog/grpc-web-interceptor.md
chalin: > You can drop this example

Done

PR #287, opened 2020-06-16
Add an image for grpc-web blog

Jennnnny:


chalin:
Just curious: what did you use to create the drawing?

> image

A few comments and suggestions concerning the image:

- Move the drawing elements as close as possible to the top-left corner.
- Ensure that the image size closely fits the drawing elements -- i.e., that there isn't a lot of whitespace.
- The interceptor labels: there is no need for two copies of the same label in each box. Drop one label and then center it in the box
- Make the arrow heads bigger.
- Your diagram seems to use a box to denote a processing element. Maybe remove the box border around "Request", "Response" etc. and instead simply place these labels just above the line (as shown in the hand sketch I shared with you earlier).
- Capitalization: "Intercepted Request" -> "Intercepted request"; similarly for the "Intercepted Response".

Thanks

Jennnnny:
Thanks for your advice! Uploaded a new file.
The diagram was drawn using google docs.

PR #309, opened 2020-06-23
Update grpc-java release tag to version 1.30.2

sanjaypujare:


chalin:
@ejona86 @creamsoup @voidzcy: regarding https://github.com/grpc/grpc.io/issues/317#issuecomment-649130939, maybe we shouldn't update the release tag (like in this PR) until all Java repositories have been updated?

I imaging that it can be disconcerting to new users who are trying out the quick start to have it fail just because the Java repos aren't up-to-date.

WDYT?

ejona86:
I think instead we should maybe split Android to using its own version of grpc-java, so it can be updated separately. There is no upper bound for the amount of time we may have to wait for jcenter. We have waited months in the past.

sanjaypujare:
I'll create another PR to change it back to 1.30.1 if JCenter doesn't get updated today.

PR #320, opened 2020-06-25
Android Java quick-start updates

chalin:
A first pass on updating prose and code that is wrong / outdated.

(Until we decide whether or not we'd like to fold the relevant parts of this page into the Java quick start, I figured that we might as well ensure that this quick start actually works -- this PR takes some steps in that direction.)


chalin:
cc @voidzcy @creamsoup

PR #343, opened 2020-07-10
Fix list item spacing

### LINKED ISSUE #305: Fix list item spacing
- [ ] Fix list item spacing (context: https://github.com/grpc/grpc.io/pull/302#discussion_r443727194)
  For another example, see https://grpc.io/docs/languages/go/quickstart/#prerequisites:
  > image

- [ ] Clean up the custom styling introduced in #302

chalin:
Fixes #305

### Screenshots

Compare the following with the screenshot in #305:

> image

The `/docs` page even looks better now (we removed the in-page custom styling):

> image

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -197,6 +197,11 @@ $colors: mergeColorMaps(("secondary": ($secondary, $white), "twitter-blue": ($tw
   .highlight
     margin-bottom: 1rem
 
+  li >
+    p:last-child:not(:only-child),
+    ul:last-child:not(:only-child)
+      margin-bottom: 1em
+
   p a
     font-weight: $weight-medium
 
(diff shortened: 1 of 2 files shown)


assets/sass/style.sass
```diff
@@ -197,6 +197,11 @@ $colors: mergeColorMaps(("secondary": ($secondary, $white), "twitter-blue": ($tw
   .highlight
     margin-bottom: 1rem
 
+  li >
+    p:last-child:not(:only-child),
+    ul:last-child:not(:only-child)
```
chalin, line 202: Note: In addition to addressing the two problems mentioned in #305, by using `:not(:only-child)` we ensure that lists with entries that contain a single (`p` or `ul`) child remain compact.

Issue #363, opened 2020-07-20
Quick Start and Tutorial page titles should include the language


chalin:
Contributes to #362

Issue #362, opened 2020-07-20
Analytics and SEO improvements


chalin:
- [x] `sitemap.xml`: some pages don't have the following entries:
  - [x] `lastmod`
  - [→] `changefreq` #571
- [x] Add breadcrumbs for at least the language pages - #233
- [x] Make page titles unique - #363
    Not strictly required, but there is some SEO weirdness that is happening with the old base Quick-start page being automatically redirect to the Go Quick-start page. It might help if at least the Quick-start and Tutorial pages mentioned the language in the title.
- [x] examples/README: update links into grpc.io Java pages https://github.com/grpc/grpc-java/pull/7288
- [x] Add `google-site-verification` file #365 & #366
- [→] Consider using campaigns -- e.g., see [URL builders](https://support.google.com/analytics/answer/1033867), if only for page access from source code repositories #572

mjpitz:
So Hugo has some built-in semantics for handling breadcrumbs. Here's a thread describing it: https://discourse.gohugo.io/t/snippet-partial-for-creating-breadcrumbs-uses-bootstrap-4/15426

Here's it being used in the Docsy theme:
https://github.com/google/docsy/blob/master/layouts/partials/breadcrumb.html

mjpitz:
I put up a PR that adds `date` to most of the content based on their last modification in git. Not sure how to configure `changefreq`. First time I'm seeing that one.

mjpitz:
Looks like changefreq should be as easy as setting some configuration in `config.yaml`: https://gohugo.io/templates/sitemap-template/

chalin:
Closed by #561
(Will track the two remaining items via their own issues.)

PR #364, opened 2020-07-20
Change non-homepage HTML titles to be "Page-title &ndash; gRPC"

chalin:
- This follows a common convention used, for example, on https://kubernetes.io.
- Prep work for #363, and contribution to #362.

Preview: https://deploy-preview-364--grpc-io.netlify.app/

cc @ejona86

### DIFF
--- a/layouts/blog/single.html
+++ b/layouts/blog/single.html
@@ -1,5 +1,5 @@
 {{ define "title" -}}
-{{ .Title }} &ndash; Blog | {{ site.Title -}}
+{{ .Title }} &ndash; {{ site.Title -}}
 {{ end -}}
 
 {{ define "main" -}}
--- a/layouts/blog/single.html
+++ b/layouts/blog/single.html
@@ -1,8 +1,8 @@
-{{ define "title" }}
-The gRPC Blog &ndash; {{ .Title }}
-{{ end }}
+{{ define "title" -}}
+{{ .Title }} &ndash; Blog | {{ site.Title -}}
+{{ end -}}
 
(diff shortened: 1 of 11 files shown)


layouts/blog/single.html
```diff
@@ -1,8 +1,8 @@
-{{ define "title" }}
-The gRPC Blog &ndash; {{ .Title }}
-{{ end }}
+{{ define "title" -}}
+{{ .Title }} &ndash; Blog | {{ site.Title -}}
```
chalin, line 2: Note the little variance in the title for blog entries: "page-title - Blog | gRPC"

chalin: I've removed "Blog"

chalin:
Following @ejona86's feedback, we're now using `&ndash;` consistently as the separator.

chalin:
@ejona86 - thanks for the feedback

ejona86:
Some reasoning: I pointed out that Gmail and Github both use the same separator multiple times. I wanted to keep " - " simply because that's what we were using already (to avoid bikeshedding and flip/flopping every year). Chrome's window itself uses a " - " on my machine (not for tabs, but the window), which is just another instance of using the same separator multiple times. In all cases though it seems the site name is _last_ in the URL, not first like we had previously; the specific page is first so that it is visible (instead of all your tabs saying "gRPC - Q..."

Aside: I noticed that '-' seems to be much more common at the moment than '|', at least in the sites I had open at the time (but that basically means "various Google properties" "Github" "Wikipedia"... I learned something about myself).

PR #366, opened 2020-07-20
Move google-site-verification file to /static

chalin:
Hmm, contrary to what is stated in [this comment](https://discourse.gohugo.io/t/add-raw-html-page-as-post/15434/4), it seems that Hugo _still_ applies a layout to an HTML file without any front matter. So, I'm moving the file to `/static` -- and double checking that it is served as is both locally, and in the preview deploy.

- Followup to #365
- Contributes to #362

cc @ejona86


chalin:
Preview check: https://deploy-preview-366--grpc-io.netlify.app/google835082a209a5e072.html
The file is deployed as is.

PR #371, opened 2020-07-27
Banner update: Join us today at gRPC conf

chalin:
- Change leading banner text to read "Join us *today*"
- Change link to general conf page; instead link directly to schedule.

cc @thisisnotapril @ejona86

### Screenshot

>


chalin:
Given that the conference is starting soon, I'll merge this now. You can still let me know if you'd like adjustments.

chalin:
Thanks for the feedback.

PR #318, opened 2020-06-25
Revert grpc-java release tag to version 1.30.1 because of Jcenter delays

sanjaypujare:
fixes #317 (by undoing #309).

attn: @chalin

PR #374, opened 2020-07-30
Update grpc-java release tag to v1.31.0

ericgribkoff:


chalin:
We had an issue earlier due to the latest release of grpc-java not being available from JCenter, see:

- [Android Java quick start: "Could not find any version that matches `io.grpc:grpc-core:[1.30.2]`" #317](https://github.com/grpc/grpc.io/issues/317)
- [Revert grpc-java release tag to version 1.30.1 because of Jcenter delays #318](https://github.com/grpc/grpc.io/pull/318), in particular https://github.com/grpc/grpc.io/pull/318#pullrequestreview-437702621 in which @ejona86 wrote:
  > We can do this for now as we figure out our longer-term approach.

  I'm not sure that we've figured out a longer-term approach yet. Comments @ejona86?

chalin:
I see that 1.30.2 made it to JCenter (more quickly than some of us might have expected 😄): https://bintray.com/bintray/jcenter/io.grpc%3Agrpc-core

chalin:
I'd vote to merge this and address the Android-Java release number issue separately.

My current thinking is that we can add a note to the Android Java pages (quick start, and tutorial) to say that folks should use the latest release available from JCenter, if the advertised release hasn't shown up yet. I think that this is preferable to creation a separate "Android Java grpc-java release version" parameter, especially since we are still hoping to merge the Android-Java pages into the (regular) Java pages.

Thoughts @ejona86?

ejona86:
@chalin, no, it still doesn't have it entirely. The problem is that the list of versions available to maven does not include it: https://jcenter.bintray.com/io/grpc/grpc-all/maven-metadata.xml . But if you look at other modules it does have 1.30.2: https://jcenter.bintray.com/io/grpc/grpc-core/maven-metadata.xml

Now, it happens that those version lists aren't always needed. They are only needed when we use version ranges. We only use version ranges when referencing grpc-core, grpc-api, and grpc-netty-shaded. So if those three have an up-to-date maven-metadata.xml, then we could probably safely use jcenter. It happens that those three do have 1.30.2.

If you look at grpc-testing-proto you can see that it isn't a simple syncing issue; it has 1.30.1 but not 1.30.0. https://jcenter.bintray.com/io/grpc/grpc-testing-proto/maven-metadata.xml

chalin:
See https://github.com/grpc/grpc-java/issues/7276.

srini100:
@ejona86, should the java release process be modified to not file this version bump PR until the artifacts are available?

ejona86:
https://github.com/grpc/grpc-java/issues/7276 is the current plan. If I end up doing a patch release on Friday I will make sure that gets in before I do so. So I don't think there is any need to update the release instructions.

ejona86:
Closing in favor of #382

PR #378, opened 2020-08-04
Return to helloworld directory

qgervacio:
Running the `protoc..` will throw `No such file or directory` since working directory is moved outside `helloworld`. Proposing to return back to `helloworld` directory after `go install .`


qgervacio:
> ...
> Did you try the instructions as given, and did they fail to work?
> ...

My bad. I copied the cmd between ()

chalin:
No problem. Thanks for confirming.

PR #382, opened 2020-08-13
Bump grpc-java to 1.31.1

ejona86:
This replaces #374, as it uses `mavenCentral()` with the Android projects so we are no longer concerned whether JCenter is out-of-date.


chalin:
Approved. There are conflicts to be resolved -- I'll let you handle that. (Let me know if you need me to do the merge.)

PR #410, opened 2020-09-22
Basic cleanup of testimonials

chalin:
Trimming file: keep only 6 entries. This is meant to be a quick cleanup. (More updates to the site will be coming in subsequent PRs, e.g., to add dev stories.)

- Drop Carbon3D: I can't find any public evidence that they're using gRPC
- Drop University of Wisconsin: not clear that the Open lambda project is still active
- Add dates for all remaining entries


ejona86:
Can you do the reformatting as a separate PR? There's far too much noise here.

> I can't find any public evidence that they're using gRPC

Is that a requirement? Do we doubt they are using gRPC or just it's not helpful if there isn't a public statement?

chalin:
> Can you do the reformatting as a separate PR? There's far too much noise here.

Yeah, I actually had it as separate commits, but then just merged them. I'll submit the reformatting separately.

chalin:
Diff should be easier to manage now. @ejona86, PTAL

chalin:
>> I can't find any public evidence that they're using gRPC
>
> Is that a requirement? Do we doubt they are using gRPC or just it's not helpful if there isn't a public statement?

Good questions.

Though my preliminary research I've found several more companies with interesting, recent and public gRPC adoption stories. As for the homepage: I would favor highlighting a small number (4-6) of companies that have dev stories that can help site readers in their choice of either adopting gRPC, or even in the steps that were taken as part of the process.

Otherwise, what is the added value for current and future developers to mention Carbon3D? As an OSS project, I think that we should favor public gRPC success stories.

ejona86:
> Otherwise, what is the added value for current and future developers to mention Carbon3D?

Most of the "mass of logos" I see around don't include links or details of any kind. Their only goal is to communicate "people are using this thing" and ideally "people like you are using this thing."

> As an OSS project, I think that we should favor public gRPC success stories.

I agree, with everything else being equal, not because we are an OSS project but because it lets people see how it is used and so it more informative. But we _know_ different languages/ecosystems skew differently on their representation in OSS. C++ in particular will have substantially limited representation. We don't want to say "gRPC is great for your OSS project." We want to say "gRPC is great for your project."

> I would favor highlighting a small number (4-6) of companies that have dev stories that can help site readers in their choice of either adopting gRPC, or even in the steps that were taken as part of the process.

That sounds great, as long as we aren't blind to representation (language and "vertical", in a "type of system" sense not market sense).

ejona86 (approved):

chalin:
Thanks for your thoughtful comments (👍), and the review.

PR #422, opened 2020-09-26
Fix PATH order in C++ quickstart guide

elelel:
Current Quickstart guide assumes complicated dependencies setup. Under this setup OS-installed packages may have gRPC binaries that conflict with the ones built with the guide's instructions. To ensure that the guide's binaries take precedence over OS-wide, PATH environment variable should be set accordingly. In general PATH variable elements have left order precedence. 

The build issue with grpc_cpp_plugin was experienced and solved by changing the PATH order on Archlinux (package probuf  3.12.4-1, protobuf-c 1.3.3-2).

### DIFF
--- a/content/docs/languages/cpp/quickstart.md
+++ b/content/docs/languages/cpp/quickstart.md
@@ -29,7 +29,7 @@ $ mkdir -p $MY_INSTALL_DIR
 Add the local `bin` folder to your path variable, for example:
 
 ```sh
-$ export PATH="$PATH:$MY_INSTALL_DIR/bin"
+$ export PATH="$MY_INSTALL_DIR/bin:$PATH"
 ```
 
 ### Prerequisites


chalin:
Thanks for the PR, and I understand your use case, but modifying the PATH variable can be tricky and we aren't trying to cover all the cases. The current PATH modification command that is given in the guide is only _one_ way to update the PATH, of course, but it is a safe way. Your change would place a `$MY_INSTALL_DIR/bin` ahead of `/usr/sbin` etc. We don't want to endorse that.

That being said, if this happens often enough for our C++ users, I suppose that we could add a warning about possible conflicts with system binaries. 

@jtattermusch @ejona86 @srini100 - comments?

ejona86:
Given we say things like "Under Linux, the version of the system-wide cmake can be too low. You can install a more recent version into your local installation directory as follows", with the current setting of PATH it does seem broken as the setup will not override the system-wide cmake.

elelel:
I see your concerns. I think, in terms of safety, the general approach is that adding an untrusted directory to PATH regardless of  the order is potentially dangerous (see https://security.stackexchange.com/questions/117535/ordering-of-the-path-environment-variable). Do the instruction describe one-time setup of gRPC for getting helloworld compiled or a possible permanent setup? For the former, which I assume is the case, there probably could be an explanation about security implications of setting PATH (which is equally potentially dangerous in any order, especially to locations like home subdirs as suggested). 

Some kind of mentioning of the potential conflict should definitely be present on tutorial page. It's quite likely that a user when trying out gRPC for the first time tries OS-provided package first. If it can't be used to compile current version of helloworld, they  will naturally turn to gRPC sources that the example comes with, without uninstalling the system packages, which was my case. Example's CMake scripts and dependent scripts some times are capable of calling the correct version of executables from build directory, some time not. As a result, they get mixed. In my case, at least incorrect (system-wide) grpc cpp generator was called. The error manifested itself only on late linking stage. To trace it all the way up to the beginning of the tutorial one must think about such potential mixups, which is not the first thing that comes to mind. A hint that the problem might be not be about the local gRPC setup but about the fact that this setup is not of the highest priority for the build scripts could be useful.

jdowens-commit:
I need to know what happened on this issue

content/docs/languages/cpp/quickstart.md
```diff
@@ -29,7 +29,7 @@ $ mkdir -p $MY_INSTALL_DIR
 Add the local `bin` folder to your path variable, for example:
 
 ```sh
-$ export PATH="$PATH:$MY_INSTALL_DIR/bin"
+$ export PATH="$MY_INSTALL_DIR/bin:$PATH"
```
jtattermusch, line 32: @chalin  I think this change is fair enough and it is indeed required to make sure e.g. user-installed cmake will be available as expected.

jtattermusch (approved):

chalin:
Thanks for your feedback @jtattermusch.

@elelel - you'll need to:
- Sign the CLA and
- Rebase / resolve the conflicts

jdowens-commit (commented):
This is an chain to exploit and should be removed and also to commit criminal activity.

elelel:
@chalin, at the time of submitting this PR I did not realize that a minor change of swapping a couple of symbols in the line would require a signed CLA. Now I'm not using gRPC and probably never will. We can close this PR and you may make the change if you find it useful. Here I express explicitly my resignation of any possible intellectual property claims on the changes proposed by myself in this PR.

> This is an chain to exploit and should be removed and also to commit criminal activity.
@jdowens-commit I'm not sure which of the two versions you are referring to by the word "this", but the original version is indeed exploitable. In fact that what led me to submitting this PR: when following the instructions in the tutorial another grpc generator binary was called (installed by a third-party system-wide package as a dependency), and not the one that was supposed to be built and run in the described procedure. But at least it was not criminal...

chalin:
@elelel - thanks again for submitting the PR, and your patience while this got discussed and resolved. Your suggested change has been incorporated via #754.

PR #423, opened 2020-09-26
In Dart docs highlight `grpc:` in `protoc` invocation.

mraleph:
In Dart docs highlight importance of `grpc:` part of the `--dart-out` option.

### DIFF
--- a/content/docs/languages/dart/basics.md
+++ b/content/docs/languages/dart/basics.md
@@ -141,6 +141,9 @@ From the `route_guide` example directory run:
 protoc -I protos/ protos/route_guide.proto --dart_out=grpc:lib/src/generated
 ```
 
+Note the `grpc:` part of the `--dart-out` command line option, that's what
+instructs `protoc` to generate gRPC related code.
+
 Running this command generates the following files in the `lib/src/generated`
 directory under the `route_guide` example directory:
 
--- a/content/docs/languages/dart/quickstart.md
+++ b/content/docs/languages/dart/quickstart.md
@@ -144,6 +144,9 @@ $ protoc --dart_out=grpc:lib/src/generated -Iprotos protos/helloworld.proto
 You'll find the regenerated request and response classes, and client and server
 classes in the `lib/src/generated` directory.
 
+Note the `grpc:` part of the `--dart-out` command line option, that's what
+instructs `protoc` to generate gRPC related code.
+
 Now implement and call the new RPC in the server and client code, respectively.
 
 #### Update the server


mraleph:
@chalin could you review?

chalin:
Hi @mraleph! Thanks for the PR.

Can you provide us with some motivation for this addition?

It feel like a **one-off remark** in at least two respects:

- For one thing, we don't highlight the meaning of _any_ of the `protoc` command-line arguments, so why single out _part of_ only _one_ of the arguments.
- Also, we don't highlight the use of the `grpc` option in any other language's Quick start or Tutorial pages. (Hmm, actually, the `grpc` option only seems visible in the Dart pages and the [Go basics tutorial](https://grpc.io/docs/languages/go/basics/#generating-client-and-server-code)).

I see that the `grpc` option is mentioned under the [Services section of the Dart Generated Code](https://developers.google.com/protocol-buffers/docs/reference/dart-generated#service) reference (from `developers.google.com/protocol-buffers`):

> The protocol buffer compiler can be invoked with the `grpc` option (e.g. --dart_out=grpc:output_folder), in which case it will generate code to support gRPC.

(That section ends with the text "See the [gRPC Dart quick start](https://grpc.io/docs/quickstart/dart.html) guide for more details" ... which is a bit misleading. It would be more accurate to write something like: "For simple working example, see the  [gRPC quick start for Dart](https://grpc.io/docs/quickstart/dart.html)".)

If you really feel like something more needs to be said, maybe one of the following options could do?

- Add a note (in the places you suggest) with a link to the Protocol Buffers [Dart Generated Code](https://developers.google.com/protocol-buffers/docs/reference/dart-generated) reference. Or, ...
- We could (also/instead) add the [Dart Generated Code](https://developers.google.com/protocol-buffers/docs/reference/dart-generated) reference link to the [Dart language landing page](https://grpc.io/docs/languages/dart/) and side nav (at the placed indicated by the red arrow:

  > Screen Shot 2020-09-28 at 17 20 59

  Other languages have their own [**Generated-code** reference pages, like the one for Go](https://grpc.io/docs/languages/go/generated-code/):

  > 

Thoughts?

cc @kwalrath

mraleph:
We got a user confused about it (https://github.com/grpc/grpc-dart/issues/362#issuecomment-699513495).

I think it is easy to overlook and omit it unless you copy command line verbatim and/or understand what every single part of it does (which a first time user might not). Other parts of the command line are kinda self explanatory hence no comments on them, `grpc:` is the only slightly magical part (hiding in plain side no less) - you can omit it and you don't get any error, just wrong file will be generated.

I don't have strong feelings about this one way or another. If you don't think this sentence carries its weight then feel free to close this PR.

chalin:
Thanks for providing some context @mraleph.

> I think it is easy to overlook and omit it unless you copy command line verbatim and/or understand what every single part of it does

Right. The expectation is that novice users should copy verbatim.

> I don't have strong feelings about this one way or another. If you don't think this sentence carries its weight then feel free to close this PR.

Ok, thanks. Then I'll close this PR and follow up with possible additions along the lines that I mentioned earlier -- via PRs addressing these issues:

- Quick start and Basics tutorial pages: consider added link to page describing protoc cmd line opts #428
- Consider adding links to the proto buf doc Generated Code reference pages #427

PR #424, opened 2020-09-28
grpc: a documentation fix in python basics guide

khanchi97:


chalin:
Hi. Thanks for your PR, but the current text is correct. It reads:

> The generated code files are called `route_guide_pb2.py` and `route_guide_pb2_grpc.py` and contain:
> 
> - classes for the messages defined in route_guide.proto
> ...

Message types, like `message Point { ... }` are defined in `route_guide.proto`, not `route_guide_pb2*.py`.

PR #435, opened 2020-10-05
Update grpc-go version to v1.32.0

chalin:


chalin:
Actually, the parameter is only used in the Quick start, and the Greeter example only works with HEAD. Closing.

PR #448, opened 2020-10-12
Go quick start: Go plugin -> Go plugins

chalin:
There are now two separate plugins, one for messages and the other for services. This tweaks the terminology slightly to reflect that fact.

cc @menghanl


menghanl (approved):

chalin:
Thanks for the review @menghanl.

PR #499, opened 2020-11-05
Config of gRPC pkg versions: make it easier/more uniform to introduce new languages

### LINKED ISSUE #498: gRPC pkg versions: make it easier/more uniform to introduce new gRPC languages
**Dart** and **Kotlin** are in need of specifying their own gRPC package versions. Currently, gRPC package versions are declared as site parameters in the config file like this:

```yaml
  grpc_release_tag: v1.33.1
  grpc_java_release_tag: v1.33.1
  grpc_go_release_tag: v1.31.0
```

We can do better than to embed the language inside the id, using a YAML map, for example.

chalin:
As suggested in #498, switching to using a **YAML map** to specify gRPC package versions for different languages.

- Closes #498.
- There are no changes in the generated site content. (New languages will be introduced in subsequent PRs.)

If any one of y'all would like to comment and/or approve, that would be grand:
@ejona86 @dfawley @jtattermusch @srini100

### DIFF
--- a/content/faq.md
+++ b/content/faq.md
@@ -61,7 +61,7 @@ The gRPC project does not do LTS releases. Given the rolling release model above
 
 ### What is the latest gRPC Version?
 
-The latest release tag is {{< param grpc_release_tag >}}.
+The latest release tag is {{< param grpc_vers.core >}}.
 
 ### When do gRPC releases happen?
 
--- a/config.yaml
+++ b/config.yaml
@@ -8,9 +8,10 @@ params:
   repo: https://github.com/grpc/grpc.io
   branch: main
   locale: en_US
-  grpc_release_tag: v1.33.1
-  grpc_java_release_tag: v1.33.1
-  grpc_go_release_tag: v1.31.0
+  grpc_vers:
+    core: v1.33.1
+    go: v1.31.0
+    java: v1.33.1
   font_awesome_version: 5.12.1
   description: A high-performance, open source universal RPC framework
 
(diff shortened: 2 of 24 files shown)


content/faq.md
```diff
@@ -61,7 +61,7 @@ The gRPC project does not do LTS releases. Given the rolling release model above
 
 ### What is the latest gRPC Version?
 
-The latest release tag is {{< param grpc_release_tag >}}.
+The latest release tag is {{< param grpc_vers.core >}}.
```
chalin, line 64: @jtattermusch @ejona86 @thisisnotapril @mjpitz : I get the impression, and the lines above are one concrete example, that the FAQ was written when there was only one gRPC package? It doesn't make sense to ask the question **What is _the_ latest gRPC Version**. 

- Should this text simply be dropped (via another PR) from the FAQ? (How likely is it that folks come to the FAQ to find a gRPC package version?)
- More generally, maybe the FAQ should be flagged for review?

chalin: I've created #501, _FAQ: some of the prose seems out-of-date / misleading_.

chalin:
This is a site-internal change. I'm going to merge now since two other updates are hinging on this PR. Comments can be addressed in followup PRs.

content/faq.md
ejona86: There's _always_ been multiple gRPC languages with their own releases. Go and Java went public the same day C core did. I do not disagree with your confusion at all, but I'll note for minor versions (like v1.33), the languages stay someone close. I wouldn't mind it being deleted. There's much better places to find this information. We could instead link to the /releases page of github for multiple repositories if we really felt the need to have it.

ejona86 (commented):
I honestly don't see much difference here, as it is simply a different spelling of the same thing. But it's fine and doesn't hurt anything. LGTM.

chalin:
Thanks!

content/faq.md
chalin: Thanks for your feedback, I've moved your comment to #501.

Muzaffer8217: muzafferbayindir015@gmail.com
Muzaffer Bayındır 
TÜRKİYE-ANTALYA 
+905327354707

-

Issue #506, opened 2020-11-11
Add link to grpc src repo docs from language homepage


chalin:
Tech writers and some site readers have expressed the preference that **_all_ gRPC docs be under grpc.io** (#98), as opposed to there being some "low-level technical" pages under a documentation folder of grpc package source repos (e.g., https://github.com/grpc/grpc/tree/master/doc).

It seems doubtful that all such grpc source repo pages will be migrated to grpc.io. A compromise is to offer a link to the source repo docs folder from the corresponding language homepage.

Related: #345 

---

Alternative names for the links were discussed during the 2020-07-20 Docs WG meeting. Here are some of the alternative names:

- Low-level technical docs
- **Additional docs** on GitHub ← preferred name (with or w/o "on GitHub)
- Extra docs (GitHub)
- Feature docs 
- GitHub docs

Issue #481, opened 2020-10-28
Use a custom layout for language homepages


chalin:
This is infrastructure work in support of #433

- [x] Introduce new layout #482
- [x] Convert / migrate all existing language-home pages - #494

PR #510, opened 2020-11-12
C# language-home: use new layout, link to examples and docs

chalin:
Contributes to #481, #345, #506

cc @ejona86

Preview: https://deploy-preview-510--grpc-io.netlify.app/docs/languages/csharp/

### DIFF
--- a/content/docs/languages/csharp/_index.md
+++ b/content/docs/languages/csharp/_index.md
@@ -1,17 +1,18 @@
 ---
-title: C# / .NET
+title: C#
+description: The original core-library implementation of gRPC for C#
 api_path: grpc/LANG/api/Grpc.Core
+layout: prog_lang_home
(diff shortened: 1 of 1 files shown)


chalin:
Consider the following as a possible followup change: _Move .NET page to Platforms?_ #511

chalin:
@jtattermusch, note the following changes made in this PR:

- The title was simplified (yes, I recall that you had originally requested that the title be changed from "C#" to "C# / .NET", but given the proposal for #511, I think that it makes sense to simplify it to "C#").
- A description was added, and the opening paragraph ("There are two official implementations of gRPC for C#...") was dropped. IMHO, this change seems to work better with the new layout.

Let me know what you think.

jtattermusch (approved):
LGTM.

PR #514, opened 2020-11-13
Python: update supported versions to 3.5+, only

chalin:
@gnossen @lidizheng - could one of you review and approve this?

Is there a version update document for https://github.com/grpc/grpc that would ensure that updates like https://github.com/grpc/grpc/pull/19084 would be followed by a corresponding update in this repo?

### DIFF
--- a/content/docs/languages/python/quickstart.md
+++ b/content/docs/languages/python/quickstart.md
@@ -6,7 +6,7 @@ weight: 10
 
 ### Prerequisites
 
-- Python 2.7, or Python 3.4 or higher
+- Python 3.5 or higher
 - `pip` version 9.0.1 or higher
 
 If necessary, upgrade your version of `pip`:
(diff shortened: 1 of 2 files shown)


gnossen (approved):

chalin:
Thanks for the review @gnossen! Any comments on the questions in the opening comment?

gnossen:
I'm not sure what exactly you mean by "version update document". Up to this point, we've mostly used the README visible on pypi.org and metadata in the Python wheel as the source of truth for our officially supported versions. We can add some language to our release document to make sure that we also update the documentation on the website when we drop support for an interpreter version or add support for a new one.

chalin:
> We can add some language to our release document to make sure that we also update the documentation on the website when we drop support for an interpreter version or add support for a new one.

Yes, that would be grand, thanks! Btw, where is that "release document"?

gnossen:
The Python release guide is an internal document detailing the release process for gRPC Python, including several sensitive steps, like how to get access to upload credentials.

chalin:
Got it. Ok then, I'll let you add a step to that document (as mentioned above), at your discretion. Thanks!

PR #525, opened 2020-11-16
Platforms page rework; add Flutter

chalin:
- Supersedes #524
- Prep work for #491

Preview: https://deploy-preview-525--admiring-torvalds-f1ffea.netlify.app/docs/platforms/

PR #529, opened 2020-11-18
Platform tweaks

mit-mit:
Following up from https://github.com/grpc/grpc.io/pull/525, I think our preference (cc @timsneath) is to do something like this instead

Preview: https://deploy-preview-529--grpc-io.netlify.app/docs/platforms/

### DIFF
--- a/content/docs/platforms/_index.md
+++ b/content/docs/platforms/_index.md
@@ -15,9 +15,7 @@ Select a target platform to get started:
   - [Android]({{< relref "android">}})
   - iOS -- coming soon
 - [Web]({{< relref "web" >}})
-
-Or, explore the following cross-platform software development kit:
-
-- Flutter -- coming soon
+- Flutter
+  - documentation coming soon
 
 [language]: {{< relref "languages" >}}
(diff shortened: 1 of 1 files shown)


chalin:
Thanks for the suggested tweaks.

Firstly, I agree that the "docs coming soon" is clearer, thanks for that suggestion.

I would like to keep the separation between the two "kinds" of "platforms": which currently has "mobile and web" in one group and flutter in another. Can we find some wording that is acceptable to y'all that would allow us to do that?

chalin:
For example, I think that the distinction between a "target platform" vs. "development platform" seems reasonable. WDYT?

cc @ejona86 @mjpitz

chalin:
I gave this some more thought and figured that, at this point, we probably don't need two list. 
PTAL at #531 (which is a PR over this PR).

chalin (approved):

chalin:
Thanks @mit-mit and @timsneath.

ejona86:
Single list SGTM.

mit-mit:
Looks good, thanks

chalin:
Great, thanks for the feedback. Next: some actual content! 😄

Issue #527, opened 2020-11-17
Make ALTS authentication pages language-specific, and move to language sections


chalin:
(Edit: I'm making this issue ALTS specific. For the base Auth page, see #543.)

~This applies to both the:~

- ~Base [Authentication](https://grpc.io/docs/guides/auth/) page~ --> see #543
- [ALTS authentication](https://grpc.io/docs/guides/auth/alts/) 
  - [x] C++ - #528, #536
  - [x] Go - #537
  - [x] Java - #538
  - [x] Python - #539
  - [x] Drop generic ALTS page - #540

Each authentication page currently contains code for all supported languages. This proposal is to do the following:

- Create a template based each of the current authentication pages
- Use the template to create language-specific versions of each page, and to add such a page below the appropriate language section
- Once this is done for all supported languages, eliminate the "catch all" authentication page

chalin:
I'm making this issue ALTS specific. For the base Auth page issue, see #514 

Closed by #528, #536, #537, #538, #539, #540

Issue #479, opened 2020-10-27
Migrate site to use the Hugo docsy theme


chalin:
**Proposal document: [Docsy migration and site rework](https://docs.google.com/document/d/1GfbZh0YOr--01_fvGih8SuAP4FcGzgCZRvl02_OiCRM)** -- see the document for a statement of purpose and rework details.

---

Migrating to [docsy](https://github.com/google/docsy) should help address the following specific issues (note that some of these items have been carried over from #107):

- [→] Setup multi-site search, to include CNFC blog, etc #521
- [→] Analytics and SEO improvements #362
- [x] Add breadcrumbs for at least the language pages - #233
- [x] Make page titles unique - #363
- [x] Support in-page TOCs for mobile and maybe desktop too #169
- [x] Navigation sidebar cannot easily be scrolled #217
- [x] Switch to using Google search #520
- [x] Sidenav: show leaf pages under platforms #549

---

Possible prep work:

- [x] Create a `docsy` branch
- [x] Add docsy submodule to main - #228, #544
- [x] Consider: Move site pages into `content/en` #505

chalin:
It seems that **[opentelemetry.io](https://opentelemetry.io) migrated to docsy** a month ago via this initial PR: [Site Refresh #247](https://github.com/open-telemetry/opentelemetry.io/pull/247) -- here's the related issue, [opentelemetry.io #203](https://github.com/open-telemetry/opentelemetry.io/issues/203).

cc @zacharysarah @celestehorgan @nate-double-u -- FYI, I'm not sure if you were aware of this migration. I like the result. (Also, notice how clean their footer is 😄 -- an issue which I've raised for grpc.io (#456))

zacharysarah:
@chalin I didn’t know about their migration, but I agree it looks great!

austinlparker:
Thanks! If y'all have any questions about our migration let me know. It was fairly straightforward.

chalin:
Done! 🎉 (Will track the two remaining items via their own issues.)

chalin:
I've deleted the `docsy` branch, as agreed during our sync, and removed the branch from [Netlify Deploy contexts for grpc.io](https://app.netlify.com/sites/grpc-io/settings/deploys#deploy-contexts).

PR #535, opened 2020-11-20
Authentication examples for Node.js and Python

bartolootrit:
There is a lack of examples with custom headers for authentication and disabled hostname checks.

### DIFF
--- a/content/docs/guides/auth.md
+++ b/content/docs/guides/auth.md
@@ -421,6 +421,77 @@ channel = google_auth_transport_grpc.secure_authorized_channel(
 stub = helloworld_pb2.GreeterStub(channel)
 ```
 
+##### With server authentication SSL/TLS and a custom header with token
+
+Client:
+
+```python
+import grpc
+import helloworld_pb2
+
+class GrpcAuth(grpc.AuthMetadataPlugin):
+    def __init__(self, key):
+        self._key = key
+
+    def __call__(self, context, callback):
+        callback((('rpc-auth-header', self._key),), None)
+
+with open('path/to/root-cert', 'rb') as fh:
+    root_cert = fh.read()
+
+channel = grpc.secure_channel(
+    'myservice.example.com:443',
+    grpc.composite_channel_credentials(
+        grpc.ssl_channel_credentials(root_cert),
+        grpc.metadata_call_credentials(
+            GrpcAuth('access_key')
+        )
+    )
+)
+
+stub = helloworld_pb2.GreeterStub(channel)
+```
+
+Server:
+
+```python
+from concurrent import futures
+
+import grpc
+import helloworld_pb2
+
+class AuthInterceptor(grpc.ServerInterceptor):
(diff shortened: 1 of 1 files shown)


chalin:
Thanks for the contribution @bartolootrit! You'll need to agree to the CLA before we can consider this PR.

chalin:
Hi all. I've added a bunch of folks as potential reviewers. If you don't think that you're the right person to review this PR, please add someone else who you believe might be better qualified.

Note that the current "generic"/"catch-all languages" Authentication page will likely be reworked into language-specific versions, and moved into each language's section (#543) -- as was done for the ALTS page (#527). But this probably won't happen until the docsy migration is complete (#479)

cc @ejona86

content/docs/guides/auth.md
```diff
@@ -505,8 +579,17 @@ var stub = new helloworld.Greeter('localhost:50051', grpc.credentials.createInse
...
 
 ```js
-var ssl_creds = grpc.credentials.createSsl(root_certs);
-var stub = new helloworld.Greeter('myservice.example.com', ssl_creds);
+const root_cert = fs.readFileSync('path/to/root-cert');
+const ssl_creds = grpc.credentials.createSsl(root_cert);
+const stub = new helloworld.Greeter('myservice.example.com', ssl_creds);
+```
+
+##### With server authentication SSL/TLS and disabled hostname checking
```
murgatroid99, line 587: I don't see "disabled hostname checking" examples for the other languages, and I don't think we should add it. That option is intended only for testing and I would want to at least make that clearer if we are going to recommend it here.

bartolootrit: Done

murgatroid99 (approved):
The Node changes look good to me now.

content/docs/guides/auth.md
```diff
@@ -421,6 +421,76 @@ channel = google_auth_transport_grpc.secure_authorized_channel(
...
+            context.abort(grpc.StatusCode.UNAUTHENTICATED, 'Invalid key')
+
+        self._deny = grpc.unary_unary_rpc_method_handler(deny)
+
+    def intercept_service(self, continuation, handler_call_details):
+        if (
+            handler_call_details.invocation_metadata
+            and handler_call_details.invocation_metadata[0] == self._valid_metadata
+        ):
+            return continuation(handler_call_details)
```
gnossen, line 476: Nit: An explicit `else` would be nice here to aid readability, especially considering the indentation.

gnossen (approved):
Thanks for the contribution! The Python changes look good!

chalin:
Thanks for the reviews all!

content/docs/guides/auth.md
chalin: @bartolootrit - can you make this update?

bartolootrit: Done

chalin: Thanks!

PR #542, opened 2020-11-23
Drop /release

### LINKED ISSUE #502: Drop the site's release file
https://grpc.io/release yields the [grpc/grpc release](https://github.com/grpc/grpc/releases) with which the [site was configured](https://github.com/grpc/grpc.io/blob/main/config.yaml).

- Is this site feature used anywhere?
- Why single out `grpc/grpc` as opposed to, say, Java or Go?

chalin:
Closes #502

Note that I haven't included a redirect rule, since IMHO, it is probably best that any old script that made use of `https://grpc.io/release` should fail with a 404, rather than be redirected and eventually result in a 200 status. If someone really feels strongly about adding a redirect rule, then maybe https://github.com/grpc/grpc/releases might make sense.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -64,11 +64,9 @@ outputFormats:
     baseName: _redirects
   RSS:
     baseName: feed
-  RELEASE:
-    baseName: release
 
 outputs:
-  home: [HTML, REDIRECTS, RSS, RELEASE]
+  home: [HTML, REDIRECTS, RSS]
 
 # Site menus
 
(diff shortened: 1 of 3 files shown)


jtattermusch (approved):
LGTM, thanks.

jtattermusch:
I updated the internal go/grpc-release doc (removed the step to update grpc.io/release).

chalin:
I've also dropped `/release` from the old GitHub-pages hosted site, for details see https://github.com/grpc/grpc.github.io/pull/871.

PR #544, opened 2020-11-24
Refresh docsy submodule

chalin:
Contributes to #479. It seems that I had already added the docsy theme as a submodule back in May via #228.


chalin:
For the record, the deploy initially failed. I had to clear the cache (on Netlify) and restart the build/deploy:

>

chalin:
I also had to clear the production build env cache:

> image

PR #553, opened 2020-12-03
Objective-C: update macOS and iOS versions to match grpc core config

chalin:
Match verions of https://github.com/grpc/grpc/pull/24282:

> Supported platform changes:
> 
> * iOS 9 or later (from 7)
> * OSX 10.10 or later (from 10.9)

@veblush @yulin-liang

### DIFF
--- a/data/officially_supported_lang_os.yaml
+++ b/data/officially_supported_lang_os.yaml
@@ -26,7 +26,7 @@
   os: [Windows, Linux, Mac]
   compilers: [Node v4+]
 - language: Objective-C
-  os: [Mac OS X 10.11+, iOS 7.0+]
+  os: [macOS 10.10+, iOS 9.0+]
   compilers: [Xcode 7.2+]
 - language: PHP
   os: [Linux, Mac]


data/officially_supported_lang_os.yaml
```diff
@@ -26,7 +26,7 @@
   os: [Windows, Linux, Mac]
   compilers: [Node v4+]
 - language: Objective-C
-  os: [Mac OS X 10.11+, iOS 7.0+]
+  os: [macOS 10.10+, iOS 9.0+]
```
chalin, line 29: @veblush @jtattermusch @yulin-liang: notice how this _used to be_ 10.11+ (rather than 10.9+). Hence, this PR downgrades the macOS version constraint to match the min version specified by https://github.com/grpc/grpc/pull/24282. Just pointing this out to ensure that it is ok.

veblush: This is weird to me because it's different from ones in the gRPC repository. I think having 10.10+ would be fine since our CI test are using that version.

chalin: > This is weird to me because it's different from ones in the gRPC repository

Just let me know which versions (for macOS and iOS) you'd like to see in the docs, and I'll make the changes. I just used 10.10+ because that is what https://github.com/grpc/grpc/pull/24282 seemed to suggest.

veblush: 10.10+ sounds reasonable to me!

veblush (approved):

chalin:
Thanks for the review @veblush

PR #567, opened 2020-12-16
Refactor Docs ToC and improve styles

aidanranney:
This second PR addresses the next batch of work needed for the Docsy migration.

These changes are meant to address the following issues:
- [x] Revises the approach to expanding and collapsing the inline ToC in the `/docs` pages - rather than template login, a CSS-driven solution was adopted. A second collapse/expand trigger has been added below the ToC
- [x] A missing Font-Awesome icon has been replaced with an equivalent (on Showcase page)
  - [ ] Missing Font-Awesome icon still present in language pages
- The background SVG on most pages has been optimized - the output was %55 percent of the original
- [x] Social links all moved to one side of footer
- [x] Fix gap in /docs sidebar ToC - did not fill view height
- Blog posts have have hard-coded images replaced with Markdown, to render images that will scale in the main content/open a modal, where compatible
- [x] Styles and class names have been revised overall to implement a namespacing scheme and reduce or improve some Bootstrap styling
  - @chalin: haven't reviewed this is full, but what I've seen looks good!

PR #576, opened 2020-12-18
Comment files which are overriding Docsy or Hugo

aidanranney:
Overrides incur a degree of technical debt as they will need to be kept up to date when the Docsy theme is updated, if there are any conflicts or new functionality which needs to be migrated. If the need for an override or piece of customization disappears, the file can be safely deleted as Hugo which default to Docsy in its lookup order.

Keeping this on draft, pending a merge and rebase of #567

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -1,3 +1,5 @@
+{{/* This file is in place as an override to the Docsy theme */}}
+
 {{/* cSpell:ignore URL's */}}
 {{ define "title" -}}
 404 Page not found &ndash; {{ site.Title -}}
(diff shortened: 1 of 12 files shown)


chalin:
> Keeping this on draft, pending a merge and rebase of #567

It has been merged, so removing draft status.

layouts/404.html
```diff
@@ -1,3 +1,5 @@
+{{/* This file is in place as an override to the Docsy theme */}}
```
chalin, line 1: Should this be just "This file is in place as an override to Docsy" like most of the other files?

chalin (approved):

layouts/404.html
aidanranney: Will put a fixup for this in next PR!

PR #577, opened 2020-12-18
Add new shortcode to add missing headings to ToC

aidanranney:
Previously, the Markdown renderer was not being passed the headings to render on "language home" list pages. This moves some of the template logic into a shortcode and specifies that it should be rendered as Markdown. The shortcode is then called directly in the index pages.

### DIFF
--- a/content/en/docs/languages/go/_index.md
+++ b/content/en/docs/languages/go/_index.md
@@ -1,7 +1,6 @@
 ---
 title: Go
 prog_lang_home: true
-notoc: true
 api_path: https://pkg.go.dev/google.golang.org/grpc
 content:
   - learn_more:
@@ -19,7 +18,9 @@ content:
 spelling: cSpell:ignore Isberner Klerk Malte youtube
 ---
 
-#### Developer stories and talks {#dev-stories}
+{{% docs/prog-lang-home-content %}}
+
+### Developer stories and talks {#dev-stories}
 
 - **[Stubbing gRPC in Go](https://jadekler.github.io/2020/10/08/stubbing-grpc.html)**,
   by [Jean de Klerk](https://github.com/jadekler), Google. October 8, 2020.
--- a/content/en/docs/languages/cpp/_index.md
+++ b/content/en/docs/languages/cpp/_index.md
@@ -1,7 +1,6 @@
 ---
 title: C++
 prog_lang_home: true
-notoc: true
 src_repo: https://github.com/grpc/grpc
 content:
   - learn_more:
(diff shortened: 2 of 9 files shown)


layouts/shortcodes/docs/prog-lang-home-content.html
```diff
@@ -0,0 +1,25 @@
+{{ $lang := lower ($.Page.Params.language | default $.Page.Params.title) }}
```
chalin, line 1: Nice solution!

Shouldn't we be able to give this file a `.md` suffix given that it is rendering markdown?

(And if so, the same would be true of any other shortcodes with a `.html` suffix that is actually being used to render markdown.)

content/en/docs/languages/go/_index.md
```diff
@@ -19,6 +18,8 @@ content:
 spelling: cSpell:ignore Isberner Klerk Malte youtube
 ---
 
+{{% docs/prog-lang-home-content %}}
+
 #### Developer stories and talks {#dev-stories}
```
chalin, line 23: - [ ] Promote heading to level 3 (to match the other headings in the rendered page like "Learn more").
- [ ] Please also make this change in all other language-landing pages that have a similar heading.

chalin (commented):
Coming along nicely! See inline comments.

layouts/shortcodes/docs/prog-lang-home-content.html
aidanranney: This won't break if you change the file type, however I'm not sure about Hugo's opinions on best practices here. I haven't been able to find much on the topic. I gave this an `.html` type though, because it's largely still concerned with the HTML structure of the lists, and because comparable shortcodes [using the `%`] in Docsy are using the `.html` type. [e.g. `blocks/feature.html`]

content/en/docs/languages/go/_index.md
aidanranney: Updated!

chalin:
- [ ] The bullets are back:
  >

layouts/shortcodes/docs/prog-lang-home-content.html
chalin: > This won't break if you change the file type, ...

Actually, since you're going to be making changes to this PR, could you change the suffix on this file to `.md`?

chalin (commented):
Looking good, pending two changes:

- Regression: the bullets are back in the subsection of language-landing pages (see PR comment for a screenshot).
- Let's try the `.md` suffix on that markdown-generating shortcode (if it works, otherwise, keep `.html` of course).

aidanranney:
Thanks! Fixed!

chalin (approved):

PR #579, opened 2020-12-19
Docsy changes/polish

aidanranney:
This moves to address the following issues (copied from the shared Google Doc):

- Docs in-page TOC: after "Content" use ">" when collapsed, and "^" when expanded

Documentation page:
- Don't show breadcrumb with single self-linking item

About & showcase on mobile:
- Currently only shows "Page Contents" w/o anything below it. (overflow bug)
- "Page Contents" → "Contents"
- Use same icons for collapsed/expanded as for doc pages
- Blog post pages should have in-page TOC too, like doc pages.

One note that is TBD is:
- Documentation page: Official support heading should be in TOC (either to be dropped, or generate a ToC with more than one heading)

### DIFF
--- a/assets/scss/_grpc.scss
+++ b/assets/scss/_grpc.scss
@@ -180,6 +180,11 @@ c - Component (Aware of its content/context...)
   }
 }
 
+.l-no-breadcrumb {
+  display: block;
+  margin-bottom: 3.25rem;
+}
+
 // Object
 .o-banner {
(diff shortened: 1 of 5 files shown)


chalin (approved):
All good except for "Currently only shows "Page Contents" w/o anything below it. (overflow bug)", which will be addressed in a followup PR.

PR #408, opened 2020-09-21
Support i18n with Korean

markruler:
Hi! 🖐

I forked this repository and started to translate Korean, and I already translated [content/ko/docs/what-is-grpc/introduction.md](https://github.com/markruler/grpc.io/blob/support-i18n/content/ko/docs/what-is-grpc/introduction.md) into Korean. I would like to commit it to community.

So, I update some files to commit.

[Add]
/i18n/...
/content/ko/...

[Update]
/config.yaml
/layouts/partials/navbar.html
/layouts/partials/navbar-lang-selector.html
/layouts/partials/searchbar.html
/layouts/partials/home/info.html
/layouts/partials/home/hero.html

PR #584, opened 2020-12-30
Add korean docs

cxsu:
Related: #408

And I recommend using `config.toml` for readability before more languages are added.


cxsu:
sign

veblush:
@cxsu Thanks for your contribution. Is it still in progress since most of contents under the kr directory are not fully translated.

@thisisnotapril @chalin Can you folllow up this PR? I don't know our strategy about having a translated doc.

lidizheng:
Ping @thisisnotapril

sergiitk:
Ping once more @thisisnotapril, @srini100.

srini100:
Thanks for the contribution. Looks like only the index files are translated. This would be a lot of effort and future maintainence burden. Also, we'll need reviewers who can understand the language. I don't think we should go this route given some browsers have pretty good translation support in many languages. I'll close this for now. Please feel free to reopen if we want to discuss this further.

PR #646, opened 2021-02-15
fix typo in state-of-grpc-web.md

JapuDCret:
Although I love steamed hams, I don't think Johan meant to write that


chalin (approved):
> Although I love steamed hams, I don't think Johan meant to write that

LOL, I'm quite sure that you're right! Thanks for catching this and submitting a PR to fix it.

PR #653, opened 2021-02-16
Updated description

Chandankkrr:
Could be one of the following
- The result was gRPC, which is now used in many organizations outside of Google 
- The result was gRPC, which is now used in many great organizations outside of Google


chalin (approved):
Thanks for the PR. I agree with your simplification.

Issue #669, opened 2021-02-22
Create and use a grpc.io-docsy theme to hold docsy file overrides


chalin:
Create a grpc-docsy theme from the docsy files that we "override" for this repo.

The proposal is to create of docsy fork (currently [chalin/docsy][]), and then use a branch (currently [chalin/docsy/tree/grpc-docsy][]) to hold all docsy files that were copied and customized for use in this site.

This will make it easier to track and integrate docsy updates.

[chalin/docsy]: https://github.com/chalin/docsy
[chalin/docsy/tree/grpc-docsy]: https://github.com/chalin/docsy/tree/grpc-docsy

Resources:

- https://www.andrewhoog.com/post/git-submodule-for-hugo-themes/ described the problem
- We might eventually be leveraging Hugo [Theme Components][], but maybe not until the docs are updated -- they currently "outdated and ... in the process of being rewritten".

[Theme Components]: https://gohugo.io/hugo-modules/theme-components

cc @nate-double-u

chalin:
This is the final home of the forked repo: https://github.com/grpc/grpc.io-docsy
The branch to be used for grpc.io is named `grpc.io`.

PR #671, opened 2021-02-22
grpc-docsy: move assets/scss to submodule

chalin:
- Contributes to #669
- There is no change in the generated site.


chalin:
FYI, @nate-double-u: this is a good example of the benefits of this approach: see, for example, the diff for `assets/scss/main.scss` since the the docsy-theme branch.

Issue #610, opened 2021-01-22
Upgrade to latest docsy version (2021/02)


chalin:
Followup to #603

chalin:
Closed by #710, #711.

PR #679, opened 2021-02-23
Build tweaks now that we're using hugo.IsProduction

chalin:
The `grpc-docy` now includes the following fix: https://github.com/google/docsy/pull/413 -- so I'm tweaking build scripts. Some more context is provided by https://github.com/gohugoio/hugo/issues/6873#issuecomment-784557603:

> `production` is the default environment you get when running `hugo`, `development` when running the server.

Contributes to #610

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,15 +9,15 @@ clean:
 
 serve:
 	@./check_hugo.sh
-	hugo server
+	hugo serve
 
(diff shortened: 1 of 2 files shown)


chalin:
I've double checked that the robots meta tag is still "index, follow", e.g.:

>

PR #691, opened 2021-03-01
C++ quick start: add note about local installs

### LINKED ISSUE #690: C++ quick start: add note about local installs
From @jtattermusch, https://github.com/grpc/grpc.io/pull/686#discussion_r584833037:

> nit: add a note that we always recommend installing under a CMAKE_INSTALL_PREFIX when using "make install" since there is not easy way to undo "make install" and it's easy to pollute your system with a global non-prefixed install?
> 
> Alternatively we can say that we strongly discourage using "make install" globally, without using a prefix.

chalin:
Closes #690


jtattermusch (approved):
LGTM. thanks

PR #694, opened 2021-03-03
Change a message name that was also a keyword

VelocityRa:
It's being picked up by the syntax highlighter the same way as all `message` keywords, so was a bit distracting.


VelocityRa:
Welp, not signing that.

PR #699, opened 2021-03-05
Write out Remote Procedure Call (RPC) in full on first use

### LINKED ISSUE #590: About: define RPC on first use
.... gRPC is a modern open source high performance RPC framework that can run in any .... 
document should spell out "RPC" on first use (i.e. what does "RPC" stand for?)

ref: https://www.grpc.io/about/#osp

chalin:
- Write out Remote Procedure Call (RPC) in full on first use, in the homepage and About page.
- Closes #590

cc @ejona86 @thisisnotapril

### DIFF
--- a/content/en/_index.md
+++ b/content/en/_index.md
@@ -42,12 +42,12 @@ Learn more
 <section class="home--top-section why-grpc">
 <h2>Why gRPC?</h2>
 
-gRPC is a modern open source high performance Remote Procedure Call (RPC) framework that can
-run in any environment. It can efficiently connect services in and
-across data centers with pluggable support for load balancing,
-tracing, health checking and authentication. It is also applicable
-in last mile of distributed computing to connect devices, mobile
-applications and browsers to backend services.
+gRPC is a modern open source high performance Remote Procedure Call (RPC)
+framework that can run in any environment. It can efficiently connect services
+in and across data centers with pluggable support for load balancing, tracing,
+health checking and authentication. It is also applicable in last mile of
+distributed computing to connect devices, mobile applications and browsers to
+backend services.
 </section>
 
 {{< home/features >}}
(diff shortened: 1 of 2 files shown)


ejona86 (approved):
Should it not be capitalized?

content/en/_index.md
```diff
@@ -42,7 +42,7 @@ Learn more
 <section class="home--top-section why-grpc">
 <h2>Why gRPC?</h2>
 
-gRPC is a modern open source high performance RPC framework that can
+gRPC is a modern open source high performance Remote Procedure Call (RPC) framework that can
```
dfawley, line 45: Any concerns about line length? The rest of this paragraph is wrapped sooner.  Same below.

chalin: I wanted to keep the diff in this PR as simple as possible.

I'll cleanup the line wrapping in followup "noop" PR -- i.e., a PR that has no effect on the generated site (as confirmed by a diff over the generated site files). Hopefully that's ok with you.

chalin:
> Should it not be capitalized?

We can drop the capitalization if we drop the "(RPC)", otherwise the convention is to capitalize -- see [Use acronyms properly](https://developers.google.com/tech-writing/one/words/#use_acronyms_properly) from the [developers.google.com/tech-writing](https://developers.google.com/tech-writing).

content/en/_index.md
chalin: Of course, I can line-wrap the paragraphs in _this_ PR. Let me know what you prefer @dfawley.

ejona86:
> We can drop the capitalization if we drop the "(RPC)", otherwise the convention is to capitalize -- see Use acronyms properly from the developers.google.com/tech-writing.

Both of those examples were names and would be ordinarily capitalized. And it didn't say explicitly to capitalize. But I can live with it either way.

content/en/_index.md
dfawley: I don't feel too strongly, but it seems better to do everything at once.

Also I would prefer the caps but, again, don't feel strongly.

chalin:
Ok, keeping the caps and push a commit that does the line wrapping.

PR #704, opened 2021-03-06
SCSS: use styles_project.scss rather than edit Docsy's main.scss

chalin:
There is no essential change to the generated site.


chalin:
cc @nate-double-u

PR #725, opened 2021-03-12
Fix small spelling mistake on core concepts, architecture and lifecycle page

jamesallured:
It looks like there was a small typo in the [Deadlines/Timeouts](https://grpc.io/docs/what-is-grpc/core-concepts/#deadlines) section on the core concepts, architecture and lifecycle page.

I did a quick check to make sure that _maynot_ wasn't a term that I hadn't heard of (in case it wasn't a mistake) but I couldn't find anything that indicated that.


chalin (approved):
Thanks for catching that and submitting a PR!

PR #739, opened 2021-03-25
Bump grpc-java for v1.36.1 release

ejona86:


yashykt (approved):

yashykt:
Would these be ideally updated as part of the release process?

ejona86:
> Would these be ideally updated as part of the release process?

What do you mean? I updated this as part of the release process. Do you mean would it be automated?

yashykt:
> What do you mean? I updated this as part of the release process. Do you mean would it be automated?

I just noticed that Core and Go were still mentioning 1.35.0 and hence my comment.

ejona86:
I see. I don't know about the other languages, but y'all have updated the versions in the past as part of the release. I don't know where you document your release process (nor Go for that matter), but we've had it documented in Java's release process since v1.1: https://github.com/grpc/grpc-java/commit/5204e4f29e28432d5fe3189b70445a7d9079820d

PR #741, opened 2021-03-29
Update basics.md

xinnad:
I got the following error when using the grpc repository version recommended in the documentation. I pulled origin/master and everything worked fine. Not sure which version to suggest but I don't think it's v1.35.0. :)

```sh
node ./dynamic_codegen/route_guide/route_guide_server.js --db_path=./dynamic_codegen/route_guide/route_guide_db.json
/home/dan/Sandbox/grpc/examples/node/node_modules/@grpc/grpc-js/build/src/server.js:72
        throw new Error('Not implemented. Use addService() instead');
        ^
```

### DIFF
--- a/content/en/docs/languages/node/basics.md
+++ b/content/en/docs/languages/node/basics.md
@@ -44,6 +44,14 @@ example too.
 To download the example, clone the `grpc` repository by running the following
 command:
 
+Not sure what version should be here, but I got the following error running the Node RouteGuide example with v1.35.0:
+```sh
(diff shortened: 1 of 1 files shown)


chalin:
As @sanjaypujare pointed out, you're reporting an issue (and thanks for that), so I'll close this PR, because the change to the documentation isn't acceptable.

Maybe #750 has addressed part of your problem. If you feel like there might still be a problem with Node grpc, consider [opening an issue](https://github.com/grpc/grpc.io/issues/new/choose).

PR #750, opened 2021-04-19
docs: node quickstart.md

ycjcl868:
`npm install` should in the `examples/node` not `examples/node/dynamic_codegen`

image


murgatroid99 (approved):
running `npm install` in the `dynamic_codegen` works for me, but this change won't break anything, so LGTM.

PR #751, opened 2021-04-20
Make CNCF shortcode a bit more generic

chalin:
... to facilitate reuse. This change will also do the right thing if/when grpc.io goes [multilingual](https://gohugo.io/content-management/multilingual/).

Generated site files are not changed by this PR.


Edward601:
> ... to facilitate reuse. This change will also do the right thing if/when grpc.io goes [multilingual](https://gohugo.io/content-management/multilingual/).
> 
> Generated site files are not changed by this PR.

> How can I retrieve everything, Link back to real data

Edward601:
@timburks I want to return to my real data & I no longer want to clone anything what shall I do

PR #756, opened 2021-04-22
Dart 2.12+ is required

chalin:
Since [grpc-dart v3.0.0](https://github.com/grpc/grpc-dart/blob/master/CHANGELOG.md#300), Dart 2.12+ is required.

cc @kevmoo


kevmoo:
LGTM!

PR #762, opened 2021-04-27
Blog: C# future -- normalizing markdown

chalin:
/cc @jtattermusch -- this PR _only_ normalizes the markdown text so that it will be easier to examine diffs in subsequent PRs. There is no change in content.

Preview: https://deploy-preview-762--grpc-io.netlify.app/blog/grpc-csharp-future/


jtattermusch:
SGTM.

PR #766, opened 2021-04-30
Publish the grpc-csharp-future blogpost

jtattermusch:
(remove the draft flag)

**NOTE**: do NOT merge before Monday, May 3.

### DIFF
--- a/content/en/blog/grpc-csharp-future.md
+++ b/content/en/blog/grpc-csharp-future.md
@@ -1,5 +1,4 @@
 ---
-draft: true
 title: The future of gRPC in C# belongs to grpc-dotnet
 date: 2021-05-01
 spelling: cSpell:ignore dotnetcore nuget Tattermusch


content/en/blog/grpc-csharp-future.md
```diff
@@ -1,5 +1,4 @@
 ---
-draft: true
 title: The future of gRPC in C# belongs to grpc-dotnet
 date: 2021-05-01
```
chalin, line 3: Will you be publishing this today? If so, we need to change the date to today (since the website doesn't currently publish future-dated blog entires).

jtattermusch: No I was planning to publish on Monday.

chalin (approved):

content/en/blog/grpc-csharp-future.md
chalin: FYI, I've added a note to that effect in the opening comment.

PR #772, opened 2021-05-05
Fix typo in go basics doc

niya3:


chalin:
The option we want is actually `--go_opt`, see [Protocol Buffers > Go Generated Code : Compiler Invocation](https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation).

niya3:
ok, this is the same as in https://github.com/grpc/grpc.io/pull/773#issuecomment-832622673

PR #773, opened 2021-05-05
Fix typo in go quickstart doc

niya3:


chalin:
Hi. Did you try the `protoc` command as it is given and it failed to work?

The option we want is actually `--go_opt`, see [Protocol Buffers > Go Generated Code : Compiler Invocation](https://developers.google.com/protocol-buffers/docs/reference/go-generated#invocation).

niya3:
@chalin thanks for noting it! Perhaps i've tried an older version of `protoc` or `protoc-gen-go` binary which was missing this option so i got the `Error - Unknown flag: --go_opt` message. In current environment i can't reproduce it, but i do remember the similar issue on SO: https://stackoverflow.com/questions/62909028/unknown-flag-go-opt-while-regenerating-grpc-code

chalin:
Thanks for clarifying, and the link to the SO question. As you mention, the issue you saw was probably because `protoc` needed to be updated to a more recent version.

PR #785, opened 2021-05-16
Fix Go quickstart docs to use local package

TakuKaneda:
When I tried the go quickstart, I found that I needed to change the import statements to use the local package.

### DIFF
--- a/content/en/docs/languages/go/quickstart.md
+++ b/content/en/docs/languages/go/quickstart.md
@@ -155,6 +155,15 @@ func (s *server) SayHelloAgain(ctx context.Context, in *pb.HelloRequest) (*pb.He
 }
 ```
 
+Also you need to fix imports in order to use the local package which you created earlier:
+
+```diff
++ pb "helloworld/helloworld"
++
+  "google.golang.org/grpc"
+- pb "google.golang.org/grpc/examples/helloworld/helloworld"
+```
+
 #### Update the client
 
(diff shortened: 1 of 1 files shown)


content/en/docs/languages/go/quickstart.md
```diff
@@ -155,6 +155,15 @@ func (s *server) SayHelloAgain(ctx context.Context, in *pb.HelloRequest) (*pb.He
 }
 ```
 
+Also you need to fix imports in order to use the local package which you created earlier:
+
+```diff
++ pb "helloworld/helloworld"
++
+  "google.golang.org/grpc"
+- pb "google.golang.org/grpc/examples/helloworld/helloworld"
```
andreybokhanko, line 164: This should work as in the original due to presence of `grpc-go/examples/go.mod` file that starts with:
```
module google.golang.org/grpc/examples
```

Not sure why it doesn't work for you; perhaps you moved this specific example out of `grpc-go/examples/`?

(though I agree with you that usage of the absolute path is confusing here)

TakuKaneda: > Not sure why it doesn't work for you; perhaps you moved this specific example out of `grpc-go/examples/`?

Indeed, I copied `grpc-go/examples/helloworld/` dir out side of `go.mod`

Thanks for the checking. I will close the PR

PR #790, opened 2021-05-25
Add performance best practices guide

ananda1066:
Ported performance best practices guide from prod to open-source per @jtattermusch's request.

### DIFF
--- a/content/en/docs/guides/performance.md
+++ b/content/en/docs/guides/performance.md
@@ -33,13 +33,14 @@ description: A user guide of both general and language-specific best practices t
     1.  **Create a separate channel for each area of high load** in the
         application, or
 
-    2.  **Use a pool of gRPC channels** to randomly distribute RPCs over
+    2.  **Use a pool of gRPC channels** to distribute RPCs over
         multiple connections (channels must have different channel args to
         prevent re-use so define a use-specific channel arg such as channel
         number). \
-        ***Side note:*** *The gRPC team has plans to add a feature to fix these
-        performance issues, so any solution involving creating multiple channels
-        is a temporary workaround that should eventually not be needed.*
+
+    ***Side note:*** *The gRPC team has plans to add a feature to fix these
+    performance issues, so any solution involving creating multiple channels
+    is a temporary workaround that should eventually not be needed.*
 
 ### C++
 
@@ -59,16 +60,10 @@ description: A user guide of both general and language-specific best practices t
     continuously getting stuck in a slow path that results in essentially serial
     request processing. 
 
-*   **Enable write batching in streams** if message k + 1 does not rely on
+*   *(Special topic)* **Enable write batching in streams** if message k + 1 does not rely on
     responses from message k by passing a WriteOptions argument to Write with
     buffer_hint set - `stream_writer->Write(message,
     WriteOptions().set_buffer_hint());`
-
-*   *(Special topic)* In the case that your application has unary or server-side
-    streaming methods where there will be exactly one request message from the
-    client, the **synchronous-with-deferred-read interface** could reduce memory
-    usage on busy servers. Read more about the interface
-    [here](https://g3doc.corp.google.com/net/grpc/g3doc/cpp/codegen.md#synchronous-with-deferred-read-interface).
 
 *   *(Special topic)*
     [gRPC::GenericStub](https://grpc.github.io/grpc/cpp/grpcpp_2generic_2generic__stub_8h.html)
@@ -83,13 +78,13 @@ description: A user guide of both general and language-specific best practices t
 
 *   **Use non-blocking stubs** to parallelize RPCs.
 
-*   **Provide a custom executor based on the workload** (cached (default), fixed, forkjoin, etc).
+*   **Provide a custom executor that limits the number of threads, based on your workload** (cached (default), fixed, forkjoin, etc).
 
 ### Python
 
 *   Streaming RPCs create extra threads for receiving and possibly sending the
     messages, which makes **streaming RPCs much slower than unary RPCs** in
-    Python.
+    gRPC Python, unlike the other languages supported by gRPC.
 
 *   **Using [asyncio](https://grpc.github.io/grpc/python/grpc_asyncio.html)** could improve performance.
 
--- a/content/en/docs/guides/performance.md
+++ b/content/en/docs/guides/performance.md
@@ -91,7 +91,7 @@ description: A user guide of both general and language-specific best practices t
     messages, which makes **streaming RPCs much slower than unary RPCs** in
     Python.
 
-*   **Using asyncio** could improve performance: https://grpc.github.io/grpc/python/grpc_asyncio.html.
+*   **Using [asyncio](https://grpc.github.io/grpc/python/grpc_asyncio.html)** could improve performance.
 
 *   Using the future API in the sync stack results in the creation of an extra
     thread. **Avoid the future API** if possible.
--- a/content/en/docs/guides/performance.md
+++ b/content/en/docs/guides/performance.md
@@ -16,11 +16,14 @@ description: A user guide of both general and language-specific best practices t
     server-to-client, or in both directions. Streams can avoid continuous RPC initiation,
     which includes connection load balancing at the client-side, starting a new
     HTTP/2 request at the transport layer, and invoking a user-defined method
-    handler on the server side. Streams, however, cannot be load
-    balanced once they have started and can be hard to debug for stream
-    failures. They also might increase performance at a small scale but can reduce 
-    scalability due to load balancing and complexity, so they should only be used 
-    when the intent is a long-lived flow of data.  \
+    handler on the server side. 
+
+    Streams, however, cannot be load balanced once they have started and can be hard 
+    to debug for stream failures. They also might increase performance at a small scale 
+    but can reduce scalability due to load balancing and complexity, so they should 
+    only be used when they provide substantial performance or simplicity benefit to 
+    application logic. Use streams to optimize the application, not gRPC.  
+
     ***Side note:*** *This does not apply to Python (see Python section for
     details).*
 
@@ -33,7 +36,7 @@ description: A user guide of both general and language-specific best practices t
     solutions:
 
     1.  **Create a separate channel for each area of high load** in the
-        application, or
+        application.
 
     2.  **Use a pool of gRPC channels** to distribute RPCs over
         multiple connections (channels must have different channel args to
@@ -66,8 +69,8 @@ description: A user guide of both general and language-specific best practices t
 
 *   *(Special topic)* **Enable write batching in streams** if message k + 1 does not rely on
     responses from message k by passing a WriteOptions argument to Write with
-    buffer_hint set - `stream_writer->Write(message,
-    WriteOptions().set_buffer_hint());`
+    buffer_hint set: \
+    `stream_writer->Write(message, WriteOptions().set_buffer_hint());`
 
 *   *(Special topic)*
     [gRPC::GenericStub](https://grpc.github.io/grpc/cpp/grpcpp_2generic_2generic__stub_8h.html)
--- a/content/en/docs/guides/performance.md
+++ b/content/en/docs/guides/performance.md
@@ -11,18 +11,20 @@ description: A user guide of both general and language-specific best practices t
     inactivity to allow initial RPCs to be made quickly without a delay (i.e.
     C++ channel arg GRPC_ARG_KEEPALIVE_TIME_MS). 
 
-*   **Use streaming RPCs** when wanting to avoid continuous RPC initiation,
+*   **Use streaming RPCs** when handling
+    a long-lived logical flow of data from the client-to-server,
+    server-to-client, or in both directions. Streams can avoid continuous RPC initiation,
     which includes connection load balancing at the client-side, starting a new
     HTTP/2 request at the transport layer, and invoking a user-defined method
-    handler on the server side. Streams are particularly well-suited to handling
-    a long-lived logical flow of data from the client-to-server,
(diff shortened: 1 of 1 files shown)


content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+        performance issues, so any solution involving creating multiple channels
+        is a temporary workaround that should eventually not be needed.*
+
+### C++
+
+*   **Do not use Sync API for performance sensitive servers.** If performance
+    and/or resource consumption are not concerns, use the Sync API as it is the
+    simplest to implement for low-QPS services.
+
+*   **Favor callback API over other APIs for all RPCs**, given that the
```
ananda1066, line 50: @vjpai Would this apply to the OSS callback API as well?

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+*   **Enable write batching in streams** if message k + 1 does not rely on
+    responses from message k by passing a WriteOptions argument to Write with
+    buffer_hint set - `stream_writer->Write(message,
+    WriteOptions().set_buffer_hint());`
+
+*   *(Special topic)* In the case that your application has unary or server-side
+    streaming methods where there will be exactly one request message from the
+    client, the **synchronous-with-deferred-read interface** could reduce memory
+    usage on busy servers. Read more about the interface
+    [here](https://g3doc.corp.google.com/net/grpc/g3doc/cpp/codegen.md#synchronous-with-deferred-read-interface).
```
ananda1066, line 71: @jtattermusch I wasn't able to find a link for the sync-with-deferred-read interface in github, any idea what I could link here?

srini100:
Thanks for this PR! Added a few reviewers for different languages.

gnossen (approved):

content/en/docs/guides/performance.md
markdroth: I don't think there is currently anything in OSS about this.  But I'd really love to see our generated code guide moved to OSS, since it's incredibly useful (for a lot more than just this one topic).

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+    solutions:
+
+    1.  **Create a separate channel for each area of high load** in the
+        application, or
+
+    2.  **Use a pool of gRPC channels** to randomly distribute RPCs over
+        multiple connections (channels must have different channel args to
+        prevent re-use so define a use-specific channel arg such as channel
+        number). \
+        ***Side note:*** *The gRPC team has plans to add a feature to fix these
```
markdroth, line 40: The formatting isn't quite right here.  This note applies to this entire bullet, not just to the second work-around option.

Stepping back for a moment, though, I'm not sure it actually makes sense to cover this topic in a performance doc, since it's really more of a missing feature.  Most of the time when people run into this, it's not about performance, it's about things that just aren't working.  And I'd really rather not tell people to create multiple channels, even though we know that they sometimes need to do so today.

If we do have to mention this, let's at least point to https://github.com/grpc/grpc/issues/21386.

markdroth (changes requested):

content/en/docs/guides/performance.md
vjpai: Yes. It's slower than sync at OSS for the time being but we  (well, you) will improve performance and it's only slower for truly high-QPS workloads at the current moment.

vjpai: ... and it's actually a usable API unlike CQ-based async

content/en/docs/guides/performance.md
vjpai: Maybe just omit the line for now and port the guide over from Google3 in a follow-on PR? I think we've never had an actual user for this interface that we know of. It was a feature request but the project that wanted it ended up going nowhere.

content/en/docs/guides/performance.md
vjpai: @markdroth how about if this moves to an advanced topic or special topic section?

vjpai: ... and of course with a caveat that the issue is a temporary problem, this is a temporary workaround, and that it won't be needed indefinitely in the future.

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+
+*   If having to use the async completion-queue API, the **best scalability
+    trade-off is having numcpu’s threads and one completion queue per thread**.
+
+*   For the async completion-queue API, make sure to **register enough server
+    requests for the desired level of concurrency** to avoid the server
+    continuously getting stuck in a slow path that results in essentially serial
+    request processing. 
+
+*   **Enable write batching in streams** if message k + 1 does not rely on
```
vjpai, line 62: I feel like this is a special topic also.

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+
+*   **Use non-blocking stubs** to parallelize RPCs.
+
+*   **Provide a custom executor based on the workload** (cached (default), fixed, forkjoin, etc).
+
+### Python
+
+*   Streaming RPCs create extra threads for receiving and possibly sending the
+    messages, which makes **streaming RPCs much slower than unary RPCs** in
+    Python.
```
vjpai, line 92: Maybe add a line reminding the reader that  this is unlke other languages.
```suggestion
    gRPC Python, unlike the other languages supported by gRPC.
```

vjpai (approved):
I already reviewed internally and have only minor comments at this point.

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+
+### General
+
+*   Always **re-use stubs and channels** when possible.
+
+*   **Use keepalive pings** to keep HTTP/2 connections alive during periods of
+    inactivity to allow initial RPCs to be made quickly without a delay (i.e.
+    C++ channel arg GRPC_ARG_KEEPALIVE_TIME_MS). 
+
+*   **Use streaming RPCs** when wanting to avoid continuous RPC initiation,
```
ejona86, line 14: I admit to people that steams are faster than unary RPCs, but I _discourage_ using them for performance reasons. You have be seriously hurting for performance before the costs make sense. If the unary RPCs would be _easier_ as a stream, go for it. But using a stream _just_ for the gRPC-level performance seems a mistake.

I _do_ tell people to use streams if it makes _their application_ faster, e.g., because of reduced synchronization between backends.

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+    active RPCs on the connection reaches this limit, additional RPCs are queued
+    in the client and must wait for active RPCs to finish before they are sent.
+    Applications with high load or long-lived streaming RPCs might see
+    performance issues because of this queueing. There are two possible
+    solutions:
+
+    1.  **Create a separate channel for each area of high load** in the
+        application, or
+
+    2.  **Use a pool of gRPC channels** to randomly distribute RPCs over
```
ejona86, line 36: s/randomly//. Round robin is common and isn't random. I know that isn't what this is actually meaning, but using random here may be misleading and doesn't seem to add value.

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,102 @@
...
+    send **raw gRPC::ByteBuffer as data** rather than serializing from some
+    proto. This can also be helpful if the same data is being sent multiple
+    times, with one explicit proto-to-ByteBuffer serialization followed by
+    multiple ByteBuffer sends. 
+
+### Java
+
+*   **Use non-blocking stubs** to parallelize RPCs.
+
+*   **Provide a custom executor based on the workload** (cached (default), fixed, forkjoin, etc).
```
ejona86, line 86: The main thing here is that the executor would have a limit on number of threads to avoid excessive context switches and memory overhead under high load. So I'd say, "Provide a custom executor that limits the number of threads, based on your workload."

content/en/docs/guides/performance.md
ananda1066: Hm interesting, this bullet does mention that "Streams are particularly well-suited to handling    a long-lived logical flow of data from the client-to-server,    server-to-client, or in both directions". We could emphasize this point at the start of this bullet as the main reason to use streaming RPCs?

content/en/docs/guides/performance.md
ananda1066: I understand this not being so much a performance issue.. I'm fine with omitting it if you have a strong opinion @markdroth. Alternatively, I can make it an advanced topic with a link to grpc/grpc#21386. Let me know which you would prefer.

content/en/docs/guides/performance.md
ananda1066: @vjpai High-QPS meaning what (a rough estimate is fine)? It seems for the OSS version, we don't want to publicize that the callback API should be favored for all RPCs at this point

content/en/docs/guides/performance.md
ananda1066: I don't have bandwidth to move the code guide over at the moment, but I'll create bugs for this internally for when I can get to this. In the meantime, I'll omit this topic as it seems fairly low-priority.

ananda1066 (commented):
Addressed the comments - PTAL and follow up where needed.

jtattermusch (approved):
Thanks for adding this! I'm happy with the contents as long as language owners are happy with the corresponding sections.

content/en/docs/guides/performance.md
ejona86: Yeah, that would be fine. Mainly I'd want to downplay per-RPC overhead.

We can mention that streams have less overall overhead, but it'd be important to downplay it and mention that abusing streams might increase performance at small scale but reduce scalability due to LB issues and complexity (although obviously some rare cases are appropriate to pay those costs). Basically, I'd focus on optimizing the application architecture.

dfawley:
What is this PR waiting for? Approval from @ejona86 and @markdroth ?

content/en/docs/guides/performance.md
```diff
@@ -33,13 +33,14 @@ description: A user guide of both general and language-specific best practices t
     1.  **Create a separate channel for each area of high load** in the
         application, or
 
-    2.  **Use a pool of gRPC channels** to randomly distribute RPCs over
+    2.  **Use a pool of gRPC channels** to distribute RPCs over
         multiple connections (channels must have different channel args to
         prevent re-use so define a use-specific channel arg such as channel
         number). \
```
markdroth, line 39: Please remove the trailing backslash here.

content/en/docs/guides/performance.md
markdroth: I'm still not hearing an argument about how this has anything to do with performance, so I'm not really convinced it belongs here.  Although to be fair, my main concern is not actually about whether or not it belongs in this particular doc; it's that I'm not really comfortable recommending that people create multiple channels, regardless of which doc we do it in.  But I guess this is the only work-around for this problem that exists today, so documenting it is fair, regardless of how unhappy it makes me. :/

So I guess this is fine.  But please do add a link to grpc/grpc#21386 for more info.

ananda1066: Added link to grpc/grpc#21386.

content/en/docs/guides/performance.md
ananda1066: Done, emphasized that streams should only be used for long-lived data flows and their impact on scalability

content/en/docs/guides/performance.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Performance Best Practices
```
markdroth, line 2: This formats a little weirdly.  I think it might be better to just say:

```
# Performance Best Practices
```

markdroth (approved):

content/en/docs/guides/performance.md
ananda1066: What part of the formatting is weird? I followed the other guides formatting and using `title` is how the guide appears in the navigation menu. I don't really see anything strange when I serve it locally?

markdroth: When I did "View file" it showed up as a sort of table.  If that's expected for the structure on grpc.io, then it's fine.

ananda1066: Oh yeah that's not how it looks in the real display. I think there's some formatting logic omitted from the rendering here.

content/en/docs/guides/performance.md
```diff
@@ -11,18 +11,20 @@ description: A user guide of both general and language-specific best practices t
...
     HTTP/2 request at the transport layer, and invoking a user-defined method
-    handler on the server side. Streams are particularly well-suited to handling
-    a long-lived logical flow of data from the client-to-server,
-    server-to-client, or in both directions. Streams, however, cannot be load
+    handler on the server side. Streams, however, cannot be load
     balanced once they have started and can be hard to debug for stream
-    failures. \
+    failures. They also might increase performance at a small scale but can reduce 
+    scalability due to load balancing and complexity, so they should only be used 
+    when the intent is a long-lived flow of data.  \
```
ejona86, line 23: Streams do not need to be long-lived to provide benefit and unary can be long-lived. Like I mentioned before, I'd say:

> so they should only be used when they provide substantial performance or simplicity benefit to application logic. Use streams to optimize the application, not gRPC.

If we want to expand on that to make it more clear, we could give examples like "grouping related requests over time to reduce application processing per message" and "making sure a multi-part operation is handled by the same backend". But really, just "if your application benefits, consider it" seems best; I think examples would be better if we were saying cases that _didn't_ work well with streams.

ejona86 (approved):

content/en/docs/guides/performance.md
ananda1066: Done, thanks!

PR #794, opened 2021-06-02
Update Core version

yashykt:


chalin (approved):
For reference, here is the link to the release: https://github.com/grpc/grpc/releases/tag/v1.38.0

Issue #808, opened 2021-07-30
Quick start - can only use path@version syntax with 'go get'


Kaszanas:
The issue surfaced when attempting to perform the steps that are outlined in: https://grpc.io/docs/languages/go/quickstart/#prerequisites

During the installation of prerequisites following the instructions in order to get gRPC working with Golang when using the following command:

```
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.26
```

I was met with an error as follows:
```
package google.golang.org/protobuf/cmd/protoc-gen-go@v1.26: can only use path@version syntax with 'go get'
```

After performing the ```go get google.golang.org/protobuf/cmd/protoc-gen-go@v1.26``` command as suggested by the language tooling the issue went away. The same issue exists for the second installation command in the instructions:

```
go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1
```

I am not sure if this is language version specific or not. The language version I use is: ```go version go1.15.1 windows/amd64```
I will send a PR with the change applied.

chalin:
Related: #682, #708 -- which change from `get` to `install`.

chalin:
And thanks for submitting the issue @Kaszanas!

Kaszanas:
This then seems to be language version-specific and one might think about including the cut-off point as to which language versions should use which commands.

This in order would present the problem of docs getting bloated, and as You only support the three latest major versions of Golang it seems like this issue is unnecessary.

I do not require any further assistance as I have resolved this issue on my system.

chalin:
The supported releases of Go are 1.14, 1.15, and 1.16. So you are in that range. I'd be curious to hear what @dfawley has to say about this. I'll reopen the issue and let Doug close it if he believes there isn't any action to take.

dfawley:
Re: get/install: https://golang.org/doc/go-get-install-deprecation

Looks like Go 1.16 added support for `@version`: https://golang.org/doc/go1.16#go-command, so we probably stick with `get` at least until 1.18 is released and 1.16 is the oldest version we support (unless the Go team breaks `go get` support before then - it will be functional in 1.17 at least).

Kaszanas:
Seems then that this could be resolved by effectively introducing two commands for installation and mentioning which versions of Golang require which command to be run. Example below:

# Installation
Depending on Your current Golang version please use one of the commands below in order to install the library.

## Using Golang < 1.15
[...]

PR #809, opened 2021-07-30
Changed "go install" to "go get" (#808)

Kaszanas:
Fixing an error with installing prerequisites where using ```go install``` failed with the following information:

```
package google.golang.org/protobuf/cmd/protoc-gen-go@v1.26: can only use path@version syntax with 'go get'
```


chalin (commented):
Thanks for the PR, but we just recently changed from `get` to `install`. For details, see https://github.com/grpc/grpc.io/issues/808#issuecomment-889836567. I'll let @dfawley comment further.

Kaszanas:
Thank You for the response @chalin
Seems to be language version specific. I am closing this PR then ;)

PR #813, opened 2021-08-09
Drop redundant postcss.config.js file

chalin:
The Docsy config file (of the same name) is used. I checked that there is no change in the development or production versions of the generated website files.

/cc @nate-double-u


nate-double-u (commented):
it's a touch late, but /lgtm :)

Issue #562, opened 2020-12-14
Add htmltest to validate links for generated site.


mjpitz:
[htmltest](https://github.com/wjdp/htmltest) is a common tool used to check/verify links on statically generated content. The is often run with two configurations. The first verifies links within the confines of the domain. The second verifies links outside of the domain.

Example internal configuration:

```
DirectoryPath: public
IgnoreDirectoryMissingTrailingSlash: true
CheckExternal: false
IgnoreAltMissing: true
```

Example external configuration:

```
DirectoryPath: public
IgnoreDirectoryMissingTrailingSlash: true
IgnoreAltMissing: true
IgnoreURLs:
  - ^https://grpc.io/.*$
```

PR #830, opened 2021-08-19
Greatly simplify build and build instructions

chalin:
- Eliminate the need to manually install Hugo since the right version is now fetched using npm.
- No need to worry about fetching submodules, this is also handled via npm (though differently).
- This PR also takes some initial **steps** towards the implementation of **link checking** (#562)

(FYI @ejona86 @gnossen)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,40 +1,20 @@
-DRAFT_ARGS = --buildDrafts --buildFuture
-BUILD_ARGS = --minify
-ifeq (draft, $(or $(findstring draft,$(HEAD)),$(findstring draft,$(BRANCH))))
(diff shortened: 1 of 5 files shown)


ejona86 (approved):
I hate `curl | bash` with a passion, but it was preexisting :smie:. Reduced usage on make as a glorified getopt is also nice.

chalin:
Thanks for the review @ejona86.

> I hate `curl | bash` with a passion, but it was preexisting

Same here, which is why the Makefile supports using `htmltext` if it is already in your path. 

> Reduced usage on make as a glorified getopt is also nice.

👍  More reductions are coming, once we have a better link-checker solution in place.

gnossen (approved):

PR #834, opened 2021-08-19
fix bidirectional streaming PHP example

lebedevsergey:
Related to this issue

Seems that there's a bug in documentation here

foreach ($notes as $n) {
  $route_note = new Routeguide\RouteNote();
  $call->write($route_note);
}
$call->writesDone();`
This code doesn't use $n variable in any way, it just sends an empty RouteNote


nimf:
@stanley-cheung can you take a look, please?

stanley-cheung (approved):
LGTM

PR #857, opened 2021-09-17
Render-link hook: support a no-icon class

chalin:
No change in generated site files. (This is in preparation for another PR.)

### DIFF
--- a/layouts/_default/_markup/render-link.html
+++ b/layouts/_default/_markup/render-link.html
@@ -1,11 +1,14 @@
 {{ $isRedirected := findRE "^(|../|/docs/(languages|platforms)/.+?/)(api|daily-builds)(/.*)?(\\?.*)?$" .Destination -}}
 {{ $isExternal := or (hasPrefix .Destination "http") $isRedirected -}}
+{{/* Until Hugo supports hook params (https://github.com/gohugoio/hugo/issues/6670), we'll inspect .Text. */ -}}
+{{ $noExternalIcon := in .Text "hk-no-external-icon" -}}
 <a href="{{ .Destination | safeURL }}"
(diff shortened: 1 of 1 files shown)


layouts/_default/_markup/render-link.html
```diff
@@ -1,11 +1,14 @@
 {{ $isRedirected := findRE "^(|../|/docs/(languages|platforms)/.+?/)(api|daily-builds)(/.*)?(\\?.*)?$" .Destination -}}
 {{ $isExternal := or (hasPrefix .Destination "http") $isRedirected -}}
+{{/* Until Hugo supports hook params (https://github.com/gohugoio/hugo/issues/6670), we'll inspect .Text. */ -}}
+{{ $noExternalIcon := in .Text "hk-no-external-icon" -}}
```
chalin, line 4: Of course, we could inspect `.Title`, but the use of `.Text` fits our current use case (as will become evident in the followup PR).

layouts/_default/_markup/render-link.html
```diff
@@ -1,11 +1,14 @@
...
 >
   {{- .Text | safeHTML -}}
-  {{ if $isExternal -}}
+  {{ if and $isExternal (not $noExternalIcon) -}}
   <i class="fas fa-external-link-alt"></i>
   {{- end -}}
-</a>
\ No newline at end of file
+</a>
+{{- /* This directive ensures that all trailing whitespace is trimmed. */ -}}
```
chalin, line 14: Adding this because it makes the hook less brittle. Otherwise it's easy to copy this file and forget to trim whitespace from the end of the file -- e.g., https://github.com/notaryproject/notaryproject.dev/pull/40. Adding a directive like this makes it obvious for all.

PR #752, opened 2021-04-20
Add abseil installation to C++ doc

veblush:
Fixes https://github.com/grpc/grpc/issues/25949.

Preview: https://deploy-preview-752--grpc-io.netlify.app/docs/languages/cpp/quickstart/

PR #870, opened 2021-09-22
Removed abseil installation from C++ doc

veblush:
gRPC can install Abseil when being installed by https://github.com/grpc/grpc/pull/26802 so a separate absel installation step isn't required anymore. This PR effectively undoes https://github.com/grpc/grpc.io/pull/752.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -25,7 +25,7 @@ params:
   branch: main
   locale: en_US
   grpc_vers:
-    core: v1.38.0
+    core: v1.40.0
     go: v1.35.0
     java: v1.40.1
   font_awesome_version: 5.12.1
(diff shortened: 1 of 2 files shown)


veblush:
@srini100 PTAL

srini100 (approved):

cdc6d:
Hi, I think building and installing abseil separately is still required for gRPC v1.40.0!  I just tried following the updated [C++ Quick Start instructions](https://grpc.io/docs/languages/cpp/quickstart/), and the hello_world example wouldn't compile for me because of missing abseil headers.  Once I built and installed abseil separately, the hello_world example built and ran.  Note that the gRPC change you mentioned, grpc/grpc#26802, is not included in the 1.40.0 tag.

veblush:
@cdc6d Yeah, you're right. I just updated it to use 1.41.0 to address this. Thanks for the report!

PR #910, opened 2021-12-20
Updated protoc code generation command

quaintdev:
Took me an hour to  figure out what is the right command. The command on the document fails with error code 255.


quaintdev:
Closing this as this is not really an issue but an error on my part.

PR #914, opened 2021-12-22
auth.md: Convert Java examples to ChannelCredentials API

ejona86:
The Google example could have used GoogleDefaultChannelCredentials, but
it could have also been using GoogleDefaultChannelBuilder years ago. I
did the minimal change of preserving the existing behavior, just
swapping to the new ChannelCredentials API.

CC @sergiitk 

This not suggesting the new API noticed when discussing https://github.com/grpc/grpc-java/issues/8772 with Sergii. That issue is really quite different, so this was more of a rabbit trail.

### DIFF
--- a/content/en/docs/guides/auth.md
+++ b/content/en/docs/guides/auth.md
@@ -502,8 +502,8 @@ server.start()
 ##### Base case - no encryption or authentication
 
 ```java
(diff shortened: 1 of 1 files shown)


dapengzhang0 (approved):
LGTM

yashykt (approved):

PR #959, opened 2022-03-29
Quick start, Python: update doc follow with example code

mrnonz:
I've already updated this example code to prevent confusion.

Follow with this section https://github.com/grpc/grpc/blob/master/examples/python/helloworld/greeter_client.py#L29


gnossen:
Great! Thanks for the contribution. I can't believe we missed this for so long.

gnossen (approved):

gnossen:
@chalin Please merge. I do not have the permissions to do so.

ejona86 (approved):

PR #962, opened 2022-04-01
Fix broken FIN Handshake article page

huseyinbabal:
https://www.tcpipguide.com/free/t_TCPConnectionTermination-2.htm page is no more exist. I suggest to use http://www.tcpipguide.com/free/t_TCPConnectionTermination-2.htm if it is not a concern to refer to an http web page.


apolcyn:
@thisisnotapril can you review? (I can't review approve/merge)

chalin (approved):
Thanks for catching that, and submitting a fix.

PR #964, opened 2022-04-04
Update Go quickstart to suggest using latest version

karanssh:
Updated the Go quickstart guide to suggest installation of latest version of `protoc-gen-go` and `protoc-gen-go-grpc`. Also added the versioned info as an option.

---

Edit(@chalin)

Preview: https://deploy-preview-964--grpc-io.netlify.app/docs/languages/go/quickstart/

### DIFF
--- a/content/en/docs/languages/go/quickstart.md
+++ b/content/en/docs/languages/go/quickstart.md
@@ -21,6 +21,12 @@ spelling: cSpell:ignore Fatalf GOPATH
    1. Install the protocol compiler plugins for Go using the following commands:
 
       ```sh
+      $ go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+      $ go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+      ```
+      or, you can fetch a particular version of the plugin using:
+      
+      ```sh
       $ go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.26
       $ go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1
       ```
--- a/content/en/docs/languages/go/quickstart.md
+++ b/content/en/docs/languages/go/quickstart.md
@@ -20,16 +20,12 @@ spelling: cSpell:ignore Fatalf GOPATH
 
    1. Install the protocol compiler plugins for Go using the following commands:
 
-      ```sh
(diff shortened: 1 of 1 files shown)


dfawley (commented):
These examples expect certain versions to be used.  While it might be reasonable to update the versions here to the latest (1.28 and 1.2), I don't think we should recommend using `@latest`, personally.

What prompted this change?

karanssh:
> What prompted this change?

I did want to update the version to the latest ones, but I thought adding a note about `@latest` might be helpful to some people? I can make it such that we have only the latest version in the `@version_number` field

dfawley:
> I did want to update the version to the latest ones, but I thought adding a note about @latest might be helpful to some people? I can make it such that we have only the latest version in the @version_number field

I would definitely prefer to keep this guide air-tight, and allowing the version of these tools to fluctuate without our control can potentially result in something that doesn't function properly for the user anymore.  I think it's gratuitous to explain to users how to use Go itself, but if you want to add separate a note explaining that the user can change the version numbers to the string "latest" in order to get the most recent versions, I would be okay with it.  If you do that, please also mention that this guide was written assuming the versions indicated, and not the latest versions.

karanssh:
@dfawley , I think that's very reasonable, I updated the PR to just use the latest versions instead, since you mentioned those are fine. I think `@latest` or `@version` in a separate note might be outside the scope of this doc. Is this fine?

karanssh:
I did run basic greeter server and compiled a proto file into it's generated code - but I didnt run a full test suite if that's what you are asking

dfawley:
> I did run basic greeter server and compiled a proto file into it's generated code - but I didnt run a full test suite if that's what you are asking

That should be fine.  The example doesn't directly reference the generated code or anything, so I think as long as the basic regnerate + run works fine, then we're good to go.  @chalin, does this LGTY?

chalin (approved):
Thanks for the updates to the versions. LGTM

Nit: **please remove the extra blank lines** that you added around the code-block

chalin:
> The example doesn't directly reference the generated code or anything, so I think as long as the basic regnerate + run works fine, then we're good to go. @chalin, does this LGTY?

Right, and yes this LGTM.

chalin:
> Nit: **please remove the extra blank lines** that you added around the code-block

Actually, never mind, I can do that cleanup for you.

chalin:
Thanks all!

PR #972, opened 2022-05-05
[DOCS] Kotlin: Add server.blockUntilShutdown() call to starting server example

Salzian:
This small PR adds a line to the Kotlin grpc docs. 

When I set up grpc-kotlin, I stumbled upon a problem were the main thread immediately quit after starting the server without any error. After I looked at the example code, I was missing a call to `blockUntilShutdown()`. I think this should be mentioned in the basic tutorial.

### DIFF
--- a/content/en/docs/languages/kotlin/basics.md
+++ b/content/en/docs/languages/kotlin/basics.md
@@ -312,6 +312,7 @@ fun main(args: Array<String>) {
   val port = 8980
   val server = RouteGuideServer(port)
   server.start()
+  server.blockUntilShutdown()
   /* ... */
 }
 ```


ejona86:
@jamesward, we should probably have a discussion about how best to assign stuff like this to Kotlin folks instead of assigning it to one of you "randomly." This looks fine from a Java perspective, and this API is a Java API, but it is only fair y'all get to take a look.

jamesward:
Thanks @ejona86 - Feel free to assign them to me and then I can delegate / reassign as necessary.  That is probably the easiest.  And this LGTM

jamesward (approved):

ejona86 (approved):

ejona86:
@Salzian, thank you!

PR #983, opened 2022-06-10
Limit Concurrent Processes of Initial Build to 8

### LINKED ISSUE #982: C++ Compilation Instructions in Quickstart Guide will Crash Most Machines
The default instructions indicate to build gRPC using the command `make -j`, which is nearly guaranteed to crash most machines given a project of this size. Building easily ate through my 16 GB of RAM at about 54% complete. After my terminal session crashed twice while building, I kept an eye on my system resources the next time around, and discovered this to be the issue. Limiting the number of concurrent jobs to 8 with `make -j 8` doesn't substantially increase build times and safely keeps RAM usage to something reasonable. The line of the guide in question is here:

https://github.com/grpc/grpc.io/blob/main/content/en/docs/languages/cpp/quickstart.md?plain=1#L117

tylercasper:
Fixes #982, to recommend building with 8 concurrent process, rather than an unlimited number.

Preview: https://deploy-preview-983--grpc-io.netlify.app/docs/languages/cpp/quickstart/

### DIFF
--- a/content/en/docs/languages/cpp/quickstart.md
+++ b/content/en/docs/languages/cpp/quickstart.md
@@ -114,7 +114,7 @@ $ cmake -DgRPC_INSTALL=ON \
       -DgRPC_BUILD_TESTS=OFF \
       -DCMAKE_INSTALL_PREFIX=$MY_INSTALL_DIR \
       ../..
-$ make -j
+$ make -j 8
 $ make install
 $ popd
 ```
--- a/content/en/docs/languages/cpp/quickstart.md
+++ b/content/en/docs/languages/cpp/quickstart.md
@@ -114,7 +114,7 @@ $ cmake -DgRPC_INSTALL=ON \
       -DgRPC_BUILD_TESTS=OFF \
       -DCMAKE_INSTALL_PREFIX=$MY_INSTALL_DIR \
       ../..
-$ make -j 8
+$ make -j 4
 $ make install
 $ popd
 ```
@@ -149,7 +149,7 @@ the steps of the previous section.
     $ mkdir -p cmake/build
     $ pushd cmake/build
     $ cmake -DCMAKE_PREFIX_PATH=$MY_INSTALL_DIR ../..
-    $ make -j
+    $ make -j 4
     ```
 
     {{% alert title="Note" color="info" %}}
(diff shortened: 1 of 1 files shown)


gnossen (commented):
This number is dependent on the number of cores in the machine. With 4 cores, this change is equivalent. Maybe instead we do something based on the number of cores present on the system.

On Mac, that would involve using `sysctl -n hw.ncpu` and on Linux `nproc`.

gnossen:
CC @jtattermusch

content/en/docs/languages/cpp/quickstart.md
```diff
@@ -114,7 +114,7 @@ $ cmake -DgRPC_INSTALL=ON \
       -DgRPC_BUILD_TESTS=OFF \
       -DCMAKE_INSTALL_PREFIX=$MY_INSTALL_DIR \
       ../..
-$ make -j
+$ make -j 8
```
jtattermusch, line 117: I'm fine with setting to `make -j4` -  that should work for most people and still gives a good speedup if you have more cores available.
I think users can figure it out and adjust the parameter quite easily if needed.

Note that line 152 also has `make -j` - let's update that to -j4 for consistency as well.

chalin (commented):
I'd be inclined to leave the commands as is. I question the validity of the original issue: it may crash some machines, but it shouldn't. Limiting parallelism will negatively impact most users.

Thoughts?

I'm ok with what @jtattermusch suggested as well. Let's just try to get closure on this PR before it becomes too dated.

chalin:
(From memory, this might not be the first time that this issue is raised. Maybe Jan's suggestion of `-j4` (used consistent in all command instances makes the most sense.)

yashykt:
Maybe just add a quick note about the number being configurable

jtattermusch:
@chalin  leaving the final decision up to you (also see @yashykt 's comment).

chalin (commented):
Ok, let's agree to keep things simple and just use **`-j 4`** in both places in this page (as @jtattermusch suggested).

chalin:
@yashykt - can you make the updates?

dfawley:
Ping @yashykt

XuanWang-Amos:
Ping @yashykt

content/en/docs/languages/cpp/quickstart.md
chalin: Done. Everywhere there was a `-j` I changed it to `-j 4`. PTAL

dfawley:
C++ is not my area so I will recuse myself from approving this PR.  I was only pinging it as part of an ongoing effort to ensure PRs don't linger for too long.

jtattermusch (approved):

PR #992, opened 2022-06-24
Java/basics: fix heading level

lingbin:
The "Simple RPC" heading should be at the next level of the
"Implementing RouteGuide", just like other types of service methods,
such as "Server-side streaming RPC".


chalin (approved):
Good catch, thanks for submitting a fix!

PR #994, opened 2022-06-27
Simplify the wording - remove the word "opaque"

surfer190:
I had to look up what opaque was and the explanations were not very clear.
To simplify and make readability more obvious I changed "opaque" to "not visible".
Please verify it makes sense.


chalin (changes requested):
The word "opaque" is correct. It is used, somewhat, in the following sense: https://en.wikipedia.org/wiki/Opaque_data_type.

PR #1007, opened 2022-07-15
Correcting the message for Hello Again

jinesh90:
it prints the same response. to correct this message. please update this document like this.

### DIFF
--- a/content/en/docs/languages/python/quickstart.md
+++ b/content/en/docs/languages/python/quickstart.md
@@ -189,8 +189,8 @@ def run():
         stub = helloworld_pb2_grpc.GreeterStub(channel)
         response = stub.SayHello(helloworld_pb2.HelloRequest(name='you'))
         print("Greeter client received: " + response.message)
-        response = stub.SayHelloAgain(helloworld_pb2.HelloRequest(name='you'))
-        print("Greeter client received: " + response.message)
+        rresponse2 = stub.SayHelloAgain(helloworld_pb2.HelloRequest(name='you'))
+        print("Greeter client received: " + response2.message)
 ```
 
 #### Run!


jinesh90:
please review above request

yashykt:
I don't see why the change is needed

gnossen:
I agree with @yashykt . This change should be a no-op.

gnossen:
Can you please explain the issue you're encountering in a bit more detail?

content/en/docs/languages/python/quickstart.md
```diff
@@ -189,8 +189,8 @@ def run():
         stub = helloworld_pb2_grpc.GreeterStub(channel)
         response = stub.SayHello(helloworld_pb2.HelloRequest(name='you'))
         print("Greeter client received: " + response.message)
-        response = stub.SayHelloAgain(helloworld_pb2.HelloRequest(name='you'))
-        print("Greeter client received: " + response.message)
+        rresponse2 = stub.SayHelloAgain(helloworld_pb2.HelloRequest(name='you'))
```
gnossen, line 192: `s/rresponse2/response2/`

jinesh90: It is needed as the response object is being over written and it prints the same message again. It is confusing.

gnossen: I'm assuming this is in response to my more general comment on this PR rather than this comment pointing out the typo in the change.

Indeed. The response object is being overwritten. Overwritten with the new response from the `SayHelloAgain` method.

>  it prints the same message again
I think you should double-check that your server `SayHelloAgain` handler is returning the text you want it to.

PR #1017, opened 2022-08-07
Copyedit core-concepts.md

lorensr:


sergiitk:
@thisisnotapril

dfawley:
Ping @thisisnotapril

dfawley:
This is some minor wordsmithing but seems fine to me.  @chalin?

dfawley (approved):

content/en/docs/what-is-grpc/core-concepts.md
```diff
@@ -93,8 +93,8 @@ on the server side.
   languages, the preferred term is *client*) that implements the same methods as
   the service. The client can then just call those methods on the local object,
   wrapping the parameters for the call in the appropriate protocol buffer
-  message type - gRPC looks after sending the request(s) to the server and
-  returning the server's protocol buffer response(s).
+  message type, send the request(s) to the server, and return the server's
```
ejona86, line 96: It took many readings, but I think I understand the previous text and this change does change the meaning. The client here I think was intended to be the application, so it was "you (the application) prepare the request and grpc will handle getting that to the server and getting responses."

A stronger change is probably to say "client application" instead of just "client" and "gRPC handles" instead of "gRPC looks after." (Other wording could be fine, those are just what popped into mind.)

lorensr: Clarified to say what the method is doing internally vs what the client is doing

chalin (commented):
LGTM if @ejona86 is happy as well.

ejona86 (approved):
Thank you! That was confusing!

Issue #1020, opened 2022-08-09
Inconsistent `RouteChat` code example block in c++ Basics tutorial


Willendless:
**Issue description**

In the grpc repository, the [`RouteChat`](https://github.com/grpc/grpc/blob/master/examples/cpp/route_guide/route_guide_server.cc) method is as follows

```cpp
  Status RouteChat(ServerContext* context,
                   ServerReaderWriter<RouteNote, RouteNote>* stream) override {
    RouteNote note;
    while (stream->Read(&note)) {
      std::unique_lock<std::mutex> lock(mu_);
      for (const RouteNote& n : received_notes_) {
        if (n.location().latitude() == note.location().latitude() &&
            n.location().longitude() == note.location().longitude()) {
          stream->Write(n);
        }
      }
      received_notes_.push_back(note);
    }
    return Status::OK;
  }
```

while in the basics tutorial website, the code block is

```cpp
Status RouteChat(ServerContext* context,
                 ServerReaderWriter<RouteNote, RouteNote>* stream) override {
  std::vector<RouteNote> received_notes;
  RouteNote note;
  while (stream->Read(&note)) {
    for (const RouteNote& n : received_notes) {
      if (n.location().latitude() == note.location().latitude() &&
          n.location().longitude() == note.location().longitude()) {
        stream->Write(n);
      }
    }
    received_notes.push_back(note);
  }

  return Status::OK;
}
```

**Expected behavior**

`received_notes` should be an instance variable and guarded by a mutex lock.

**Additional info**

Both the code block and the explanation in the basics tutorial should be updated.

PR #1021, opened 2022-08-09
fix: resolve inconsistency between doc and code

Willendless:
close: #1020 

Two changes are made

1. update `RouteChat` code block
2. add explanation about why use mutex lock to guard `received_notes_` instance variable


drfloob (approved):
Thanks for the fix!

yashykt (approved):

Issue #1022, opened 2022-08-10
Inconsistent Async client code example block in c++ Basics tutorial


Willendless:
**Issue description**

In the grpc repository, the stub API used in [`GreeterClient::SayHello`](https://github.com/grpc/grpc/blob/master/examples/cpp/helloworld/greeter_async_client.cc) method is as follows

```cpp
    // stub_->PrepareAsyncSayHello() creates an RPC object, returning
    // an instance to store in "call" but does not actually start the RPC
    // Because we are using the asynchronous API, we need to hold on to
    // the "call" instance in order to get updates on the ongoing RPC.
    std::unique_ptr<ClientAsyncResponseReader<HelloReply> > rpc(
        stub_->PrepareAsyncSayHello(&context, request, &cq));
```

while in the basics tutorial website, the code block is

```cpp
CompletionQueue cq;
std::unique_ptr<ClientAsyncResponseReader<HelloReply> > rpc(
    stub_->AsyncSayHello(&context, request, &cq));
```

**Expected behavior**

Async API should be `stub_->PrepareAsyncSayHello()`.

drfloob:
Thanks for pointing this out, it's now fixed. @chalin I don't have permissions to close the issue, please help out.

PR #1023, opened 2022-08-10
fix: update async api code block

Willendless:
close: #1022 

modifications

+ update code block to use `PrepareAsyncSayHello` in Asynchronous-API tutorial


drfloob:
Thanks for this. You would also need to add `rpc->StartCall()` to the tutorial here, or else it would not issue the RPC. In this example however, the "Prepare" family of methods is not very useful since the call is immediately started afterwards, so I've changed the example code to match the tutorial instead.

PR #1051, opened 2022-09-25
Improved 'make' commandline option

ntolazzi:
The former call 'make -j' keeps spawning an unlimited amount of threads. This reproducibly leads to memory exhaustion on my system (Manjaro Linux with 32GB RAM) and a subsequent crash. 'make -j $(nproc)' only spawns a limited number of compilations threads adjusted to the available cores.


XuanWang-Amos:
Thanks for the change, can you please sign the CLA? We'll start review after that.

PR #1064, opened 2022-10-25
[doc] update more protocol buffers part

asimkt:
The example was missing from the previous doc.

The example mentioned in "You'll see an example of this below." was missing.


dfawley (approved):

dfawley:
This change seems fine to me.  @chalin, could you please confirm & merge if so?

ejona86 (approved):

PR #1079, opened 2022-12-21
Rework statement about "opacity" of metadata

### LINKED ISSUE #1005: Is metadata _really_ "opaque to gRPC"?
@ejona86 - #994 and #996 got me thinking: I question the truth of the statement that "Metadata is opaque to gRPC itself" since #996 highlights that there are restrictions on the values that keys can take on -- and hence the metadata isn't opaque.

Maybe we can just drop the mention of opaqueness from that section of the docs?

https://github.com/grpc/grpc.io/blob/2c8783c1737fdbd324a3e2d65fc0c69f91758ba9/content/en/docs/what-is-grpc/core-concepts.md?plain=1#L200

chalin:
- Closes #1005

### DIFF
--- a/content/en/docs/what-is-grpc/core-concepts.md
+++ b/content/en/docs/what-is-grpc/core-concepts.md
@@ -202,7 +202,7 @@ Keys are case insensitive and consists of ASCII letters, digits, and special cha
 `.` and must not be start with `grpc-` (which are reserved for gRPC itself).
 Binary-valued keys end in `-bin` while ASCII-valued keys do not.
 
-User-defined metadata is not used by gRPC, which allows the client provide information
+User-defined metadata is not used by gRPC, which allows the client to provide information
 associated with the call to the server and vice versa.
 
 Access to metadata is language dependent.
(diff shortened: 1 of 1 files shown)


ejona86 (approved):

content/en/docs/what-is-grpc/core-concepts.md
```diff
@@ -202,7 +202,7 @@ Keys are case insensitive and consists of ASCII letters, digits, and special cha
 `.` and must not be start with `grpc-` (which are reserved for gRPC itself).
 Binary-valued keys end in `-bin` while ASCII-valued keys do not.
 
-Metadata is opaque to gRPC itself - it lets the client provide information
+User-defined metadata is not used by gRPC, which allows the client provide information
```
dfawley, line 205: `which allows the client _to_ provide ...`

chalin: Right, thanks for the careful read @dfawley!

elidemwill83 (approved):

PR #1080, opened 2022-12-27
Update supported Python version

XuanWang-Amos:
Update supported version to `3.7` to be aligned with [`setup.py`](https://github.com/grpc/grpc/blob/master/setup.py#L549).

### DIFF
--- a/content/en/docs/languages/python/quickstart.md
+++ b/content/en/docs/languages/python/quickstart.md
@@ -7,7 +7,7 @@ weight: 10
 
 ### Prerequisites
 
-- Python 3.5 or higher
+- Python 3.7 or higher
 - `pip` version 9.0.1 or higher
 
 If necessary, upgrade your version of `pip`:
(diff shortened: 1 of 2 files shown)


XuanWang-Amos:
@gnossen PTAL

gnossen (approved):
LGTM for now, but long term, we probably want some way to keep this automatically in sync with our support range, since it will be changing every year.

By the way, how did you notice this was out of date?

XuanWang-Amos:
> LGTM for now, but long term, we probably want some way to keep this automatically in sync with our support range, since it will be changing every year.
> 
> By the way, how did you notice this was out of date?

Yeah agree that we should find a way to keep those in sync.

I was looking an issue related to supported language and os and noticed that python version listed is not correct, can't find the issue link now :(

ejona86 (approved):

Josemoe:
Thanks for letting me in

PR #1123, opened 2023-05-02
Update _index.md

larry-safran:
Actually answer the question asked in the title


larry-safran:
FAQ has an answer, so this isn't really needed.

PR #1128, opened 2023-05-11
User guide for custom load balancing policies

temawi:
This is the first new piece of content for the "Guides" section in a while. We are hoping to follow this up with several other guides that highlight the main gRPC features and link to example code.

### DIFF
--- a/content/en/docs/guides/custom-load-balancing.md
+++ b/content/en/docs/guides/custom-load-balancing.md
@@ -1,7 +1,7 @@
 ---
 title: Custom Load Balancing Policies
 description: >-
-  Explains custom load balancing policies can help optimize load balancing
+  Explains how custom load balancing policies can help optimize load balancing
   under unique circumstances.
 ---
 
@@ -19,18 +19,20 @@ to the servers and picking a connection to use when an RPC is sent.
 ### Implementing Your Own Policy
 
 By default the `pick_first` policy will be used. This policy actually does no
-load balancing but just uses the first address it gets from the name resolver.
-By updating the gRPC service config you can also switch to using `round_robin`
-that rotates through all the addresses it gets. But if the built-in load
-balancing policies do not meet your needs (the exact set available vary 
-between languages) you can also implement you own custom policy.
+load balancing but just tries each address it gets from the name resolver and
+uses the first one it can connect to. By updating the gRPC service config you
+can also switch to using `round_robin` that connects to every address it gets
+and rotates through the connected backends for each RPC. There are also some
+other load balancing policies available, but the exact set varies by language.
+If the built-in policies do not meet your needs you can also implement you own
+custom policy.
 
 This involves implementing a load balancer interface in the language you are
 using. At a high level, you will have to:
 
 - Register your implementation in the load balancer registry so that it can
 be referred to from the service config
-- Define the JSON configuration object of your implementation. This allows your
+- Parse the JSON configuration object of your implementation. This allows your
 load balancer to be configured in the service config with any arbitrary JSON
 you choose to support
 - Manage what backends to maintain a connection with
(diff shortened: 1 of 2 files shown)


temawi:
@ejona86 Roughly what you would expect?

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -0,0 +1,75 @@
...
+more servers.
+
+A gRPC load balancing policy is given a list of server IP addresses by the name
+resolver. The policy is responsible for maintaining connections (subchannels)
+to the servers and picking a connection to use when an RPC is sent.
+
+### Implementing Your Own Policy
+
+By default the `pick_first` policy will be used. This policy actually does no
+load balancing but just uses the first address it gets from the name resolver.
```
ejona86, line 22: It tries each address in order and uses the first that connects.

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -0,0 +1,75 @@
...
+By updating the gRPC service config you can also switch to using `round_robin`
+that rotates through all the addresses it gets. But if the built-in load
```
ejona86, line 24: "that connects to every address it gets and rotates through the connected backends each for each RPC"

ejona86, line 24: Add something about other load balancing policies may exist. Like we have WRR now, and there's least_request in Java. I'm not suggesting to enumerate them all here.

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -0,0 +1,75 @@
...
+balancing policies do not meet your needs (the exact set available vary 
+between languages) you can also implement you own custom policy.
+
+This involves implementing a load balancer interface in the language you are
+using. At a high level, you will have to:
+
+- Register your implementation in the load balancer registry so that it can
+be referred to from the service config
+- Define the JSON configuration object of your implementation. This allows your
```
ejona86, line 33: The definition of this configuration object is sort of virtual. I'd probably emphasize "_parsing_ the JSON configuration object of your implementation."

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Service Mesh
+
+If you have a service mesh setup where a central control plane is coordinating
+the configuration of your microservices, you cannot configure your custom load
+balancer directly via the service config. But support is provided to do this
+with the xDS protocol that your control plane uses to communicate with your
+gRPC clients.
+
+For more details, please see gRPC [proposal A52]
```
ejona86, line 64: Fix this link

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -0,0 +1,75 @@
...
+(https://github.com/grpc/proposal/blob/master/A52-xds-custom-lb-policies.md).
+
+### Language Support
+
+| Language | Example | Notes |
+| -------- | ------- | ----- |
+| Java | [link](java-example) | |
```
ejona86, line 71: Use square brackets for the link name `[link][java-example]`. "link" is a very poor link name, so you could do `[java-example][]` which is short-hand for "the text is the same as the link name". You can also have spaces just fine, so can do `[Java example][]` and then at the end have `[Java example]: THE://URL`

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Custom Load Balancing Policies
+description: >-
+  Explains custom load balancing policies can help optimize load balancing
```
larry-safran, line 4: s/Explains custom/Explains how custom/

content/en/docs/guides/custom-load-balancing.md
temawi: It was in parenthesis, but I made the point more prominent.

ejona86: Yeah, I think the issue is it needed to be before the "But"

content/en/docs/guides/custom-load-balancing.md
```diff
@@ -42,6 +44,16 @@ call path
 The exact steps vary by language, see the language support section for some
 concrete examples in your language.
 
+```mermaid
+flowchart TD
+NR(Name Resolver) -->|Provides addresses &\nLB config| LB(Load Balancer)
+LB --> |Provides a picker| C(Channel)
+C -->|Picks\na subchannel| P(Picker)
```
ejona86, line 51: Would it be better to say the channel "requests a subchannel" from the picker?

ejona86 (approved):

content/en/docs/guides/custom-load-balancing.md
temawi: Yes!

PR #1137, opened 2023-05-24
First draft of cancellation doc

gnossen:


gnossen:
@temawi

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Cancellation
+description: >-
+  Explains how and when to cancel to RPCs.
```
sanjaypujare, line 4: "to cancel RPCs"

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+---
+
+### Overview
+
+When a gRPC client is no longer interested in the result of an RPC call, it may
+_cancel_ to signal this discontinuation of interest to the server.  After that
+point, the server should stop any ongoing computation and end its side of the
+connection. Often, servers are also clients to upstream servers, so that
```
sanjaypujare, line 12: "connection" ? People might confuse this with the underlying TCP or HTTP connection

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+A(Client) -->|Cancelled| B(Server 1)
+B-->|Cancelled|C(Server 2)
+```
+
+### Cancelling an RPC Call on the Client Side
+
+A client cancels an RPC call by calling a method on the call object or, in some
+languages, on the accompanying context object. At the level of the wire
+protocol, a client cancels an RPC call by sending an
+[`RST_STREAM`](https://datatracker.ietf.org/doc/html/rfc7540#section-6.4) frame
```
sergiitk, line 43: RFC 7540 obsoleted by RFC 9113:  https://www.rfc-editor.org/rfc/rfc9113.html

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+which will be called when a cancellation is received. This callback is expected
+to coordinate with the handler thread/coroutine to stop local processing on
+behalf of the request and clean up any associated resources, including upstream
+RPCs. In other languages, the mechanism may be a multithreading Event object,
+which will be set when the RPC has been cancelled.  Regardless of the
+language-specific mechanism, the thread/coroutine processing the RPC must
+periodically check if RPC has been cancelled and if it has, cease processing and
+begin the clean-up process.
+
+```mermaid
```
XuanWang-Amos, line 67: Can we add a description for this graph?

content/en/docs/guides/cancellation.md
sergiitk: Oh, it also comes with nicer the anchors:  https://www.rfc-editor.org/rfc/rfc9113#CANCEL

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+### Cancelling an RPC Call on the Client Side
+
+A client cancels an RPC call by calling a method on the call object or, in some
+languages, on the accompanying context object. At the level of the wire
+protocol, a client cancels an RPC call by sending an
+[`RST_STREAM`](https://datatracker.ietf.org/doc/html/rfc7540#section-6.4) frame
+with an error code of
+[`CANCEL(8)`](https://datatracker.ietf.org/doc/html/rfc7540#section-7). gRPC
```
sanjaypujare, line 45: Do we need to provide this level of detail in a user doc? A gRFC or a technical doc yes but why a user doc?

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+clients do not provide additional details to the server about the reason for the
+cancellation. When the server receives this frame, the application-provided
+server handler may be busy processing the request in a thread, coroutine, or
+other unit of concurrency. The gRPC library in general does not have a mechanism
+to interrupt the server handler, so the server handler must coordinate with the
+gRPC library to ensure that local processing of the request ceases and any
+upstream RPC calls initiated in service of this RPC call are themselves
+cancelled.
+
```
larry-safran, line 54: Talk about the cancel call taking a description (even though it isn't passed to the server) which can be used for logging or throwing exceptions.

ejona86 (commented):
Deadline triggers cancellation when it fires. We should cross-mention that in the docs as appropriate.

Cancellation occurs due to I/O errors as well.

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+languages, on the accompanying context object. At the level of the wire
+protocol, a client cancels an RPC call by sending an
+[`RST_STREAM`](https://datatracker.ietf.org/doc/html/rfc7540#section-6.4) frame
+with an error code of
+[`CANCEL(8)`](https://datatracker.ietf.org/doc/html/rfc7540#section-7). gRPC
+clients do not provide additional details to the server about the reason for the
+cancellation. When the server receives this frame, the application-provided
+server handler may be busy processing the request in a thread, coroutine, or
+other unit of concurrency. The gRPC library in general does not have a mechanism
+to interrupt the server handler, so the server handler must coordinate with the
```
sanjaypujare, line 50: I think what is implied here is that the application code needs to do this. "server handler" as application code may not be obvious

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+gRPC library to ensure that local processing of the request ceases and any
+upstream RPC calls initiated in service of this RPC call are themselves
+cancelled.
+
+The exact mechanism for notifying the server handler that the RPC being handled
```
sergiitk, line 55: Asked bard to shrink :)

> The mechanism for notifying the server handler of an RPC cancellation differs between languages. Some languages use a callback function, while others use an event object. Regardless of the mechanism, the thread/coroutine processing the RPC must periodically check if it has been cancelled and, if so, cease processing and begin cleanup.

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+    subgraph Thread1
+      direction TB
+      grpc("gRPC library") -->|calls| cb("onCancel CB")
+      cb --> set("sets cancelled to true")
+    end
+    subgraph Thread2
+      direction TB
+      cancelled{cancelled?} -->|false| perform("perform some work")
+      perform --> cancelled
+      cancelled -->|true| cleanup("cancel upstream RPCs")
```
sanjaypujare, line 82: And also do other clean up like clean up resources (close file handles etc)

content/en/docs/guides/cancellation.md
```diff
@@ -0,0 +1,101 @@
...
+  subgraph Server2
+  end
+
+  Client -->|CANCEL| Server1
+  Server1 -->|CANCEL| Server2
+```
+
+### Language Support
+
+| Language | Example        | Notes                            |
```
sanjaypujare, line 96: Notes column is empty - nothing to add there?

content/en/docs/guides/cancellation.md
gnossen: Thanks for the suggestion. This section was removed for being too technical.

content/en/docs/guides/cancellation.md
gnossen: The graph has been simplified to gloss over this. The answer to that question varies from language to language.

content/en/docs/guides/cancellation.md
gnossen: Removed.

content/en/docs/guides/cancellation.md
gnossen: More instances of "application-provided" have been added.

content/en/docs/guides/cancellation.md
gnossen: Shrunk.

content/en/docs/guides/cancellation.md
gnossen: Notes added

temawi (approved):

PR #1154, opened 2023-06-22
Update Node example documentation for example migration to grpc-node repository

murgatroid99:
The actual migration is occurring in grpc/grpc-node#2474 and grpc/grpc#33522. I used `master` for the node version because at first the examples will only exist on the master branch. After the next release it should be updated to `@grpc/grpc-js@1.9.0`.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -29,7 +29,7 @@ params:
     core: v1.55.0
     go: v1.56.1
     java: v1.56.0
-    node: master
+    node: "@grpc/grpc-js@1.9.0"
   font_awesome_version: 5.12.1
   gcs_engine_id: 788f3b1ec3a111a2f
   ui:
--- a/config.yaml
+++ b/config.yaml
@@ -29,6 +29,7 @@ params:
     core: v1.55.0
     go: v1.56.1
     java: v1.56.0
+    node: master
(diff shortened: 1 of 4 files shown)


chalin (commented):
While I haven't walked through the steps from those two edited pages, but it LGTM.

A few notes:

- The **PHP quickstart** makes use of the Node quickstart. There's at least one place in the PHP quickstart where you'll need to **update a path**: https://github.com/grpc/grpc.io/blob/f9ef809c06e577a13e0444f9730b05a471195c26/content/en/docs/languages/php/quickstart.md?plain=1#L178
- Can you **create an issue** to ensure that `params.grpc_vers.node` is updated once the repo has an official release. You might want to add the URL to the issue in the config file next to the `node` entry so that we have some context.

murgatroid99:
I updated the PHP quickstart, and I also updated the node version in `config.yaml`. That tag doesn't exist quite yet, but it will very soon.

murgatroid99:
The node version tag now exists.

chalin (approved):
LGTM, thanks.

murgatroid99:
I don't have write access to this repository. This is ready to merge, whenever someone who does have write access wants to.

PR #1156, opened 2023-06-28
Show a banner with a gRPConf 2023 speaker callout

temawi:
BPvWWsoZN6txtBb

### DIFF
--- a/layouts/partials/banner.html
+++ b/layouts/partials/banner.html
@@ -1,9 +1,11 @@
 {{ if site.Params.show_banner -}}
 <div class="o-banner">
   <p>
-    <a href="https://events.linuxfoundation.org/grpc-conf/program/cfp" target="_blank" rel="noopener">
+    <a href="https://events.linuxfoundation.org/grpc-conf" target="_blank" rel="noopener">
       <strong>gRPConf 2023</strong></a>
-    is happening on Sep 20th - submit your talk proposal now!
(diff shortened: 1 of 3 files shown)


layouts/partials/banner.html
```diff
@@ -1,12 +1,9 @@
...
 <div class="o-banner">
   <p>
-    For
     <a href="https://events.linuxfoundation.org/grpc-conf/" target="_blank" rel="noopener">
-      <strong>gRPC Conf 2020</strong></a>
-    session videos, visit the
-    <a href="https://www.youtube.com/playlist?list=PLj6h78yzYM2NN72UX_fdmc5CZI-D5qfJL" target="_blank" rel="noopener">
-      CNCF YouTube channel</a>
+      <strong>gRPConf 2023</strong></a>
+    is happening on Sep 20th - submit your talk proposal now!
```
ejona86, line 6: Do we want a link to the the gRPC Conf page or the page to submit your talk?

ejona86: Oh, it was already linking to the grpconf page, because it is the same link as last time. I missed that. The old link would have been fine. You're choice.

ejona86 (approved):

layouts/partials/banner.html
temawi: I ended up including a link to the top page as well as a direct link to submit a proposal.

PR #1172, opened 2023-07-26
config.yaml: update Go version to v1.57.0

dfawley:


ejona86 (approved):
My, aren't we punctual.

PR #1174, opened 2023-07-31
Clarify what GCP platforms are supported by gRPC's ALTS libraries.

matthewstevenson88:
The current wording in the documentation has led to confusion among some users of gRPC's ALTS libraries (example: https://github.com/grpc/grpc-go/issues/6252). This PR clarifies the GCP platforms on which gRPC's ALTS libraries are supported.


matthewstevenson88:
@srini100 or @thisisnotapril: Would you mind taking a look?

Apologies, I do not have permission to set reviewers. :)

ejona86 (approved):

PR #1190, opened 2023-09-06
Metadata user guide

larry-safran:
Fixes b/275364225 as part of fixit


content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Metadata
+description: >-
+  Explains what metadata is, how it is transmitted, and what it is used for.
+---
+
+### Overview
+
+Metadata is a side channel that allows clients and servers to provide
+information to each other that is associated with the data messages.
```
ejona86, line 10: "associated with the data messages" could be misleading, making someone think it is available for each message. The "sent with a gRPC request or response" would reinforce that view.

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+
+gRPC metadata is a key-value pair of data that is sent with a gRPC request or
+response. It is used to provide additional information about the call, such as
+authentication credentials, tracing information, or custom headers.
+
+gRPC metadata is implemented using HTTP/2 headers. The keys and values are ASCII
```
ejona86, line 16: The values are ASCII strings but can be binary data reads weird. I see the similar sentence in the gRPC concepts, but it reads better there (albeit, not being clear that only values may be binary).

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+strings, but the values can also be binary data. The keys are case insensitive
+and must not start with the prefix grpc-, which is reserved for gRPC itself.
+
+gRPC metadata can be sent and received by both the client and the server. It is
+sent in the initial headers of the RPC call, and can also be sent in subsequent
```
ejona86, line 21: I don't think this is clear when these things happen. I don't know if you are referring to client-side or server-side for the initial headers, for example.

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+trailers.
+
+gRPC metadata is useful for a variety of purposes, such as:
+
+* **Authentication**: gRPC metadata can be used to send authentication credentials to
+the server. This can be used to implement different authentication schemes, such
+as OAuth2 or JWT.
```
ejona86, line 28: It might be fair to mention "using the standard HTTP Authorization header" here, to make it more concrete.

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+* **Tracing**: gRPC metadata can be used to send tracing information to the server.
+This can be used to track the progress of a request through a distributed
+system.
+* **Custom headers**: gRPC metadata can be used to send custom headers to the server.
```
ejona86, line 32: Why not from a server to a client?

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+This can be used to implement application-specific features, such as load
+balancing, rate limiting or providing detailed error messages from the server
+to the client.
+* **Internal usages**: metadata is used by gRPC to communicate how to handle an
```
ejona86, line 36: Seems this would just be better to remove. Does it help the user to know this?

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+Custom metadata must follow the "Custom-Metadata" format listed in
+[PROTOCOL-HTTP2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
+, with the exception of binary headers not needing to be base64 encoded.
+
+
+
+#### Headers
+Headers are sent before any data.  The header includes things like
```
ejona86, line 54: "before any data" makes it sound like this can't be used on server-side.

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+authentication credentials and how to handle the RPC.  
+
+Some of the headers, such as authorization are generated by gRPC for you.
```
ejona86, line 57: For some reason there is a hard line break here in the HTML view. Unclear why.

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+, with the exception of binary headers not needing to be base64 encoded.
+
+
+
+#### Headers
+Headers are sent before any data.  The header includes things like
+authentication credentials and how to handle the RPC.  
+
+Some of the headers, such as authorization are generated by gRPC for you.  
+Custom header handling is language dependent. For example, in Java and Go they
```
ejona86, line 58: It'd be fair to say interceptors are used in most languages (or just "they are generally provided through interceptors." and remove the talking about language).

content/en/docs/guides/metadata.md
```diff
@@ -0,0 +1,101 @@
...
+are generally provided through Interceptors.
+
+#### Trailers
+Trailers are a special kind of header that is sent after the message data.  They
+are used internally to communicate the outcome of an RPC or that an RPC has been
+cancelled.  At the application level, custom trailers can be used to communicate
```
ejona86, line 64: RPC cancellation does not use trailers.

ejona86 (commented):
My biggest concern is the ambiguity between client/server throughout.

content/en/docs/guides/metadata.md
larry-safran: If they are trying to look at the HTTP headers directly, it might be confusing if they didn't realize that gRPC is providing ones in addition to whatever the user defines.  I've reworded to focus on that rather than our use cases.

content/en/docs/guides/metadata.md
larry-safran: reworded

content/en/docs/guides/metadata.md
larry-safran: reworded

content/en/docs/guides/metadata.md
larry-safran: reworded

content/en/docs/guides/metadata.md
larry-safran: done

content/en/docs/guides/metadata.md
larry-safran: reworded

content/en/docs/guides/metadata.md
larry-safran: reworded

content/en/docs/guides/metadata.md
larry-safran: reworded

content/en/docs/guides/metadata.md
larry-safran: Updated

content/en/docs/guides/metadata.md
```diff
@@ -48,25 +54,25 @@ Custom metadata must follow the "Custom-Metadata" format listed in
...
-Trailers are a special kind of header that is sent after the message data.  They
-are used internally to communicate the outcome of an RPC or that an RPC has been
-cancelled.  At the application level, custom trailers can be used to communicate
+
+Trailers are a special kind of header that is sent after the message data. They
+are used internally to communicate the outcome of an RPC. At the application
+level, custom trailers can be used to communicate
 values known only after the message data has been completely generated such as
-digital signatures and integrity checks or anything else the server wants to
+digital signatures and integrity checks, or anything else the server wants to
```
ejona86, line 72: Digital signatures and integrity checks are much better served in the messages themselves. Maybe a better example would be server utilization and query cost; those we actually do with ORCA.

content/en/docs/guides/metadata.md
```diff
@@ -1,45 +1,51 @@
...
+gRPC metadata is implemented using HTTP/2 headers. The keys are ASCII
+strings, while the values can be either ASCII strings or binary data. The keys
+are case insensitive
+and must not start with the prefix `grpc-`, which is reserved for gRPC itself.
 
-gRPC metadata can be sent and received by both the client and the server. It is
-sent in the initial headers of the RPC call, and can also be sent in subsequent
-trailers.
+gRPC metadata can be sent and received by both the client and the server.
+Headers are sent from the client to the server on the initial request and from
```
ejona86, line 23: Metadata is not sent with the messages, except as an optimization for unary. For bidi, headers are simply before any messages (because there could be zero), and trailers is only on server-side after its response messages.

content/en/docs/guides/metadata.md
larry-safran: Changed into a separate paragraph.  Was caused by 2 trailing spaces.

content/en/docs/guides/metadata.md
```diff
@@ -69,10 +69,9 @@ Custom header handling is language dependent, generally through interceptors.
 
 Trailers are a special kind of header that is sent after the message data. They
 are used internally to communicate the outcome of an RPC. At the application
-level, custom trailers can be used to communicate
-values known only after the message data has been completely generated such as
-digital signatures and integrity checks, or anything else the server wants to
-communicate outside of the return values.
+level, custom trailers can be used to communicate things not directly part of
+the data, such as server utilization and query cost.  Trailers can be sent by
```
ejona86, line 73: Trailers can only be sent from server to client. Ditto above.

content/en/docs/guides/metadata.md
ejona86: This still needs updating.

larry-safran: It seems unfair not to let clients use trailers, but .... updated.

content/en/docs/guides/metadata.md
larry-safran: Updated to reflect this.

ejona86 (approved):

ejona86:
CC @temawi

PR #1191, opened 2023-09-07
User guide for health checking

temawi:


temawi:
@YifeiZhuang I couldn't add you as a reviewer, but @ejona86 mentioned this is something you know about as well.

content/en/docs/guides/health-checking.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+Note that here you can also use an empty string if you want the channel to
+react to any service on the server becoming unhealthy.
+
+Enabling health checking changes some behavior around calling a server:
+
+  - The client will additionally call the `Watch` RPC on the health check
+  service when a connection is established
+    - If the call fails, retries will be made (with exponential backoff)
```
YifeiZhuang, line 67: append: unless the call fails with status `UNIMPLEMENTED`, in which case the health checking is disabled.

content/en/docs/guides/health-checking.md
```diff
@@ -0,0 +1,108 @@
...
+  - The calls will resume if the service later becomes healthy
+  - Some load balancing policies can choose to disable health checking is
+  it the feature does not make sense with the policy (e.g. `pick_first` does
+  this)
+
+More specifically the state of the subchannel (that represents the physical
+connection to the server goes through these states based on the health of
+the service it is connecting to.
+
+```mermaid
```
YifeiZhuang, line 81: That starting point of the diagram is a dot, can we change it to make it meaningful? like IDLE?
It looks the diagram is missing another part:  READY state goes to TRANSIENT_FAILURE if unhealthy. Otherwise READY state goes to IDLE if health checking is not enabled.

temawi: These UML style state diagrams always start with that dot, but I added the IDLE state as the first state we enter. Made other updates too based on your feedback.

YifeiZhuang (approved):

content/en/docs/guides/health-checking.md
```diff
@@ -0,0 +1,115 @@
...
+```json
+{
+  "healthCheckConfig": {
+    "serviceName": "foo"
+  }
+}
+```
+
+Note that here you can also use an empty string if you want the channel to
+react to any service on the server becoming unhealthy.
```
ejona86, line 61: That is a bit misleading, as empty string doesn't combine results from the individual services.

content/en/docs/guides/health-checking.md
```diff
@@ -0,0 +1,115 @@
...
+  service when a connection is established
+    - If the call fails, retries will be made (with exponential backoff), unless
+    the call fails with the status UNIMPLEMENTED, in which case health checking
+    will be disabled.
+  - Requests won't be sent until the health check service sends a healthy
+  status for the service being called
+  - If a healthy service becomes unhealthy the client will no longer send
+  requests for that service
+  - The calls will resume if the service later becomes healthy
+  - Some load balancing policies can choose to disable health checking is
```
ejona86, line 75: "is it the feature"? "it" is probably "if"

ejona86 (approved):
I'm wary of talking about subchannel states and transitions, as some of that is changing, but that diagram does seem to have value. You may mention that a failing subchannel state allows the load balancer to choose a working subchannel instead.

content/en/docs/guides/health-checking.md
temawi: Made it more clear.

PR #1205, opened 2023-10-04
Fix typo in Kotlin quickstart

temawi:
Same changes as in #984 where we could not get the CLA signed.


dfawley (approved):
I suspect this change is correct, but it might be great if @bshaffer or @lowasser could confirm.

lowasser:
LGTM.

PR #1217, opened 2023-10-28
Fix broken link

xpume:


ejona86 (approved):
Looks like the HTML page is available at https://jeanbza.github.io/jadekler.github.io/2020/10/08/stubbing-grpc.html , but the github rendering looks more attractive at no real downfall. So I agree that is the better link.

PR #1228, opened 2023-12-08
update deprecated http2_protocol_options envoy field

lcostea:
Replace the deprecated `http2_protocol_options` field - https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-http2-protocol-options

A similar change was done in all envoy examples a while ago: https://github.com/envoyproxy/envoy/pull/15563

### DIFF
--- a/content/en/docs/platforms/web/basics.md
+++ b/content/en/docs/platforms/web/basics.md
@@ -113,7 +113,11 @@ static_resources:
   - name: echo_service
     connect_timeout: 0.25s
     type: LOGICAL_DNS
-    http2_protocol_options: {}
+    typed_extension_protocol_options:
+      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
+        explicit_http_config:
+          http2_protocol_options: {}
     lb_policy: ROUND_ROBIN
     load_assignment:
       cluster_name: echo_service


content/en/docs/platforms/web/basics.md
```diff
@@ -113,7 +113,11 @@ static_resources:
   - name: echo_service
     connect_timeout: 0.25s
     type: LOGICAL_DNS
-    http2_protocol_options: {}
+    typed_extension_protocol_options:
+      envoy.extensions.upstreams.http.v3.HttpProtocolOptions:
+        "@type": type.googleapis.com/envoy.extensions.upstreams.http.v3.HttpProtocolOptions
```
sampajano, line 118: Do you know what is the specific Envoy version where this starts to work?

sampajano (approved):
@lcostea Oh thanks so much for the change!! Appreciate it!! 😊

---

Do you care to update the few occurrences in the grpc-web repo as well below?

https://github.com/search?q=repo%3Agrpc%2Fgrpc-web+path%3Aenvoy.yaml&type=code

Totally ok if you prefer not to, in which case i can too!

Thanks again for your help!

content/en/docs/platforms/web/basics.md
lcostea: In 1.23 it works for sure (didn't test in earlier versions). And current supported versions are 1.25+

lcostea: I will update also in the grpc-web repo as well (I only searched for grpc.io for occurrences). Thanks for pointing me to it.

ejona86 (approved):

content/en/docs/platforms/web/basics.md
sampajano: > In 1.23 it works for sure (didn't test in earlier versions). And current supported versions are 1.25+

Thanks for the info!!

> I will update in the grpc-web repo as well (I only searched for grpc.io for occurrences). Thanks for pointing me to it.

@lcostea Oh that's great! Thank you so much!! 😊

PR #1240, opened 2023-12-27
Point to the monthly community meetups

larry-safran:
Add a line pointing to the monthly community meetups.


larry-safran:
@temawi When you get back can you look at and merge this?

yashykt (approved):

PR #1241, opened 2023-12-31
Remove flag.Parse() line from go basic guide

pmokeev:
It's kind of unnecessary to put here `flag.Parse()` line. I guess it became from some of existing gRPC guides, where port of server receives using command line arguments.


yashykt (approved):

yashykt:
Thanks for the change!

PR #1252, opened 2024-02-01
Update quickstart.md

guptaaashutosh:
updated to Regenerate gRPC code for node which is not mention in the docs .

### DIFF
--- a/content/en/docs/languages/node/quickstart.md
+++ b/content/en/docs/languages/node/quickstart.md
@@ -100,6 +100,36 @@ message HelloReply {
 
 Remember to save the file!
 
+### Regenerate gRPC code
+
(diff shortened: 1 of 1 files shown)


apolcyn:
cc @murgatroid99

murgatroid99:
This change is incorrect. This quickstart guide uses the "dynamic codegen" example, which uses the `@grpc/proto-loader` library to generate protobuf classes at runtime. There is no need for a code regeneration step, and any code generated this way would not be used in the example.

sergiitk:
@murgatroid99. Should this PR be closed them, or still can be fixed? cc @temawi

murgatroid99:
I think it should be closed. I don't see any need to add a section like this at all.

PR #1259, opened 2024-02-23
Banner for in-person meetup.

temawi:
image

### DIFF
--- a/layouts/partials/banner.html
+++ b/layouts/partials/banner.html
@@ -1,11 +1,11 @@
 {{ if site.Params.show_banner -}}
 <div class="o-banner">
   <p>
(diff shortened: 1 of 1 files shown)


temawi:
@larry-safran FYI

ejona86 (approved):
On narrower screens the text wraps and the banner covers part of the gRPC logo. But this won't be here that long.

temawi:
> On narrower screens the text wraps and the banner covers part of the gRPC logo. But this won't be here that long.

Yeah, this has been a bit unfortunately an issue with previous banners as well. The banner has been built to go over the page, not to push the page contents down.

PR #1265, opened 2024-02-28
Updated protoc version for gRPC 1.62

veblush:
Updated the protoc-version to have the same version that gRPC 1.62 is built with.


ejona86 (approved):
FWIW, some of the links look to be out-dated (but functional). I see one referencing protobuf in the google github org, for example.

PR #1268, opened 2024-02-29
Use absolute paths instead of absolute URLs

ejona86:
When you specify grpc.io in a URL, it is considered a separate site and gets the "linking away" anchor behavior, with the icon and opening in a new tab. Also, such links don't work properly when running locally or using the PR deploy preview.

In metadata.md, the concepts guide moved locations (see layouts/index.redirects), so let's inline that redirect.


ejona86:
And they _also_ let hugo validate the links. And now it discovered one doesn't exist. I'm going to revert my change just for that one link, but @temawi, you may want to modify service-config.md to fix the link to the retry guide (which doesn't exist yet).

temawi (approved):

PR #1273, opened 2024-03-20
Update basics.md

hanifshahy:


hanifshahy:
using firstWhere() method causes error firstWhereOrNull() should be used instead of that

ejona86 (approved):

content/en/docs/languages/dart/basics.md
```diff
@@ -290,7 +290,7 @@ Future<RouteSummary> recordRoute(
   await for (var location in request) {
     if (!timer.isRunning) timer.start();
     pointCount++;
-    final feature = featuresDb.firstWhere((f) => f.location == location,
+    final feature = featuresDb.firstWhereOrNull((f) => f.location == location,
         orElse: () => null);
```
ejona86, line 294: Actually, should this orElse be removed? That's what was done in https://github.com/grpc/grpc-dart/commit/7e2c2665cfaa18e13979f60c2f2bd8f7d2daf7bc

mosuem: Yes, `firstWhereOrNull` does not provide an `orElse` parameter, as this is null by default, see the [docs](https://api.flutter.dev/flutter/package-collection_collection/IterableExtension/firstWhereOrNull.html).

PR #1275, opened 2024-03-26
[C++] Remove mention of `set_buffer_hint`

eugeneo:
fixes: grpc/grpc#34823


yashykt (approved):
I would prefer we figure out what is wrong with this API, but I don't mind removing it from the documentation for now.

PR #1277, opened 2024-04-03
Replace link to python docs from 2 to 3

jnsp:
The link to `iterator` docs is connected to python2, not python3.


gnossen (approved):

gnossen:
Thanks for the fix!

PR #1278, opened 2024-04-03
gRPConf 2024 CFP banner

temawi:


layouts/partials/banner.html
```diff
@@ -1,11 +1,10 @@
 {{ if site.Params.show_banner -}}
 <div class="o-banner">
   <p>
-    Level up your API development with gRPC! Subscribe to the
-    <a href="https://www.youtube.com/@grpcio" target="_blank" rel="noopener">
-      <strong>gRPC YouTube channel</strong>
+    gRPConf 2024 is on Aug. 27th -
+    <a href="https://sessionize.com/grpconf2025" target="_blank" rel="noopener">
```
ejona86, line 5: Should this link work? :stuck_out_tongue:

temawi: Thanks for catching that! Somehow 2024 turned into 2025 in the link.

ejona86 (approved):

layouts/partials/banner.html
ejona86: It could be a marketing trick, like with car models.

PR #1279, opened 2024-04-03
Change gRPC-go version to 1.63

arvindbr8:
Released: https://github.com/grpc/grpc-go/releases/tag/v1.63.0


ejona86 (approved):
My, aren't we punctual.

PR #1313, opened 2024-07-02
script for gtm

Gayathri625:
Created partials to embed Google Tag Manager (GTM) script. GTM container ID is dynamically fetched from config.yaml, ensuring that tags fired are sent to their assigned GTM container

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -18,6 +18,7 @@ enableGitInfo: true
 enableRobotsTXT: true
 
 params:
+  gtmID: "GTM-55D2HXLW"
   copyright: gRPC Authors
   repo: &repo https://github.com/grpc/grpc.io
   github_repo: *repo
--- a/layouts/partials/gtm-noscript.html
+++ b/layouts/partials/gtm-noscript.html
@@ -1,0 +1,8 @@
+{{ if .Site.Params.gtmID }}
+<!-- Google Tag Manager (noscript) -->
+<noscript>
+  <iframe src="https://www.googletagmanager.com/ns.html?id={{ .Site.Params.gtmID }}"
+  height="0" width="0" style="display:none;visibility:hidden"></iframe>
+</noscript>
+<!-- End Google Tag Manager (noscript) -->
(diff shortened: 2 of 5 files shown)


temawi (approved):

temawi (commented):
Instead of modifying the theme in https://github.com/grpc/grpc.io-docsy/pull/6 could we instead add these with `layouts/partials/hooks/head-end.html` and `body-end.html`? Or is it a problem having them at the end of head/body?

Mahi-coding (approved):

likhitsomana (approved):

temawi:
> > Instead of modifying the theme in [grpc/grpc.io-docsy#6](https://github.com/grpc/grpc.io-docsy/pull/6) could we instead add these with `layouts/partials/hooks/head-end.html` and `body-end.html`? Or is it a problem having them at the end of head/body?
> 
> Yeah we can gtm.html can be placed anywhere in the head. gtm-noscript.html is to track the events in browsers with JavaScript disabled. Placing gtm-noscript.html at the end of the body might cause some events to be missed, but this won't be a major issue since users with JavaScript disabled are rare

Ok great, let's try it this way and abandon the other PR. I see you have made the changes here already, is this ready to be merged?

Gayathri625:
> > > Instead of modifying the theme in [grpc/grpc.io-docsy#6](https://github.com/grpc/grpc.io-docsy/pull/6) could we instead add these with `layouts/partials/hooks/head-end.html` and `body-end.html`? Or is it a problem having them at the end of head/body?
> > 
> > 
> > Yeah we can gtm.html can be placed anywhere in the head. gtm-noscript.html is to track the events in browsers with JavaScript disabled. Placing gtm-noscript.html at the end of the body might cause some events to be missed, but this won't be a major issue since users with JavaScript disabled are rare
> 
> Ok great, let's try it this way and abandon the other PR. I see you have made the changes here already, is this ready to be merged?

yes, it is ready to merge

PR #1320, opened 2024-07-11
Modify banner styling

Gayathri625:
modified the current banner position from fixed to absolute to ensure that it does not consume screen space while scrolling and adjusted the banner z-index so the banner moves beneath the navbar while scrolling.


assets/scss/_grpc.scss
```diff
@@ -445,3 +443,7 @@ c - Component (Aware of its content/context...)
     display: none;
   }
 }
+
+.banner-container{
+  height: 60px; 
+}
```
arjan-bal, line 449: nit: Missing empty line.

arjan-bal:
To be reviewed by @temawi who has context about these changes.

assets/scss/_grpc.scss
```diff
@@ -43,7 +43,6 @@ sections site wide to the original site's background svg.
 .td-navbar .navbar-brand span:last-child {
   display: none;
 }
-
```
arjan-bal, line 46: I don't think this was intentional? If it wasn't please revert.

assets/scss/_grpc.scss
```diff
@@ -229,10 +228,9 @@ c - Component (Aware of its content/context...)
 // Object
 .o-banner {
   width: 100%;
-  position: fixed;
-  margin-left: -15px;
+  position: absolute;
   padding-top: 15px;
-  z-index: 32;
+  z-index: $z-index-navbar - 1;
```
arjan-bal, line 233: nit: Please add a single line comment just above this line explaining why we need to ensure the z-index for the banner is less than the navbar.

assets/scss/_variables_project.scss
```diff
@@ -17,3 +17,5 @@ $display4-weight: $_font-weight-light;
 $link-color: #379c9c;
 
 $enable-rounded: false;
+
+$z-index-navbar: 32;
```
arjan-bal, line 21: We're not applying this to the navbar? The navbar could change the actual z-index to 30 and then this value would not be valid anymore. We should ensure the navbar uses this variable to set the z-index.

arjan-bal: Realised that the change to the navbar is in https://github.com/grpc/grpc.io-docsy/pull/8.

We should define the variable in the docsy theme and used it here instead of the other way around.

PR #1328, opened 2024-07-21
Modify banner styling

Gayathri625:
modified the current banner position from fixed to absolute to ensure that it does not consume screen space while scrolling and adjusted the banner z-index so the banner moves beneath the navbar while scrolling.
NOTE: [#9](https://github.com/grpc/grpc.io-docsy/pull/9) has to merged before merging this PR, as the styles defined here are applied correctly when the banner is in header section.

### DIFF
--- a/assets/scss/_grpc.scss
+++ b/assets/scss/_grpc.scss
@@ -9,11 +9,6 @@ what the workaround is for
   cursor: pointer;
 }
 
-// ensure enough space for banner without overlapping with the main section
-.banner-container { 
-  height: 60px;
-}
-
 // External link icon
 a > .fa-external-link-alt::before,
 a > span > .fa-external-link-alt::before {
@@ -239,7 +234,7 @@ c - Component (Aware of its content/context...)
 // Object
 .o-banner {
   width: 100%;
-  position: absolute;
+  position: relative;
   padding-top: 15px;
   // keep banner's z-index value less than the navbar, so that banner moves beneath the navbar when user scrolls the page
   z-index: $navbar-z-index - 1;
@@ -250,14 +245,12 @@ c - Component (Aware of its content/context...)
   & p {
     padding: 0.5rem;
   }
+  padding-bottom: 0.01rem;
 }
 
 @include media-breakpoint-down(sm){
   .o-banner{
-    top: 6rem;
-  }
-  .td-sidebar{
-    padding-top: 1rem;
+    top: 0rem;
   }
 }
 
--- a/assets/scss/_grpc.scss
+++ b/assets/scss/_grpc.scss
@@ -252,6 +252,13 @@ c - Component (Aware of its content/context...)
   }
 }
 
+// Adjust the top value for smaller screens of banner banner medium and large screens to avoid overlapping with banner
+@media (max-width: 768px), (max-width: 576px) {
+  .o-banner {
+    top: 6rem; 
+  }
+}
+
 .o-icon {
   margin-right: 0.25rem;
 }
--- a/assets/scss/_grpc.scss
+++ b/assets/scss/_grpc.scss
@@ -9,6 +9,11 @@ what the workaround is for
   cursor: pointer;
 }
 
+//ensure enough space for banner without overlapping with the main section
+.banner-container { 
+  height: 60px;
+}
+
 // External link icon
 a > .fa-external-link-alt::before,
 a > span > .fa-external-link-alt::before {
@@ -229,10 +234,9 @@ c - Component (Aware of its content/context...)
 // Object
 .o-banner {
   width: 100%;
-  position: fixed;
-  margin-left: -15px;
+  position: absolute;
   padding-top: 15px;
-  z-index: 32;
+  z-index: $navbar-z-index - 1;
   top: 4rem;
   background: $light;
   text-align: center;
--- a/assets/scss/_grpc.scss
+++ b/assets/scss/_grpc.scss
@@ -241,6 +241,7 @@ c - Component (Aware of its content/context...)
   width: 100%;
   position: absolute;
   padding-top: 15px;
+  // keep banner's z-index value less than the navbar, so that banner moves beneath the navbar when user scrolls the page
   z-index: $navbar-z-index - 1;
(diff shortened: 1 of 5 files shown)


Gayathri625:
The current banner is overlapping with main section
Screenshot 2024-07-21 9 57 49 PM
Hence creating enough space for banner by wrapping the banner in a div with sufficient height.
Screenshot 2024-07-21 10 05 45 PM

Gayathri625:
Decreasing the padding values of hero section as it occupying more screen space
Screenshot 2024-07-21 10 06 48 PM

layouts/shortcodes/page/header.html
```diff
@@ -1,13 +1,13 @@
 {{ $title := $.Page.Params.title -}}
 {{ $desc  := $.Page.Params.description | markdownify -}}
 <div class="text-left">
-  <h1 class="display-1 mb-5">
```
Gayathri625, line 4: Removed mb-5 (margin-bottom) from h1 to avoid unnecessary spacing when h3 is absent. Applied mt-5 (margin-top) to h3 instead to ensure consistent spacing between h1 and h3 only when h3 is present.
eg: On community page h3 is absent after h1, hence the spacing should not be applied.

assets/scss/_grpc.scss
```diff
@@ -44,6 +49,17 @@ sections site wide to the original site's background svg.
   display: none;
 }
 
+// Override Docsy theme padding values to reduce vertical space taken up by the hero section
```
arjan-bal, line 52: What is meant by `hero` section?
IIUC, the hero refers to the "gRPC" logo in the top left inside the navbar. The style below seems to push down the content of the `main` section so that they don't go below the `header` section (which now coantains the banner + navbar). If this is the case, we should call it the `header` section and not the `hero` section.

arjan-bal:
This seems to have broken the home page

image

assets/scss/_grpc.scss
```diff
@@ -44,6 +49,17 @@ sections site wide to the original site's background svg.
...
+.td-default main > section:first-of-type {
+  padding-top: 6rem;
+  padding-bottom: 3rem;
+}
```
arjan-bal, line 56: Is this required because of the changes in the PR or is it and independent change to make the spacing look better?

If it is the latter, I would not suggest making this as perfect spacing is subjective. I also see media queries being used set styles for this element which can interact with this.

Gayathri625:
> This seems to have broken the home page
> 
> image

No, this is bcoz the [#9](https://github.com/grpc/grpc.io-docsy/pull/9) is not merged.

assets/scss/_grpc.scss
Gayathri625: Hero Section is an Introductory Section of Content that can be placed immediately before or after the Header. In our case 
Screenshot 2024-07-22 12 26 36 PM
This is the hero section.

assets/scss/_grpc.scss
Gayathri625: As mentioned in this comment https://github.com/grpc/grpc.io/pull/1328#issuecomment-2241707712
Inorder to reduce the vertical space taken by the section we are doing this. 
I am overriding only the padding values set by media queries, remaining styles still apply.

arjan-bal:
> No, this is bcoz the [#9](https://github.com/grpc/grpc.io-docsy/pull/9) is not merged.

Please mention the dependencies as a NOTE in the PR description. It should mention which PR needs to be submitted before this and why. We don't want to put the website in a weird state while another PR is being reviewed.

assets/scss/_grpc.scss
arjan-bal: I would not recommend overriding media queries unless absolutely necessary.

The initial comment was asking if this change is absolutely necessary. If not, I would suggest not overriding this css.

arjan-bal (approved):
LGTM, note that you will also need to do a submodule update in this PR once https://github.com/grpc/grpc.io-docsy/pull/9 is merged.

Command
```
git submodule update --init --recursive
```

See https://git-scm.com/book/en/v2/Git-Tools-Submodules

gnossen (approved):

gnossen:
@Gayathri625 Is this good to merge now that https://github.com/grpc/grpc.io-docsy/pull/9 is merged?

Gayathri625:
> @Gayathri625 Is this good to merge now that [grpc/grpc.io-docsy#9](https://github.com/grpc/grpc.io-docsy/pull/9) is merged?

Yes, I have updated the submodule. This can be merged

arjan-bal (changes requested):
The styling seems to break for mobile devices. The banner is covered by the navbar which becomes taller. We should fix this before merging. Let me know if you need help with this.
image

Gayathri625:
I see that the current grpc.io website also has the same issue. It is because of the lack of breakpoints on different screens for the banner. The current top value for the banner is fixed at 4rem. While the navbar takes more vertical space on small screens, the top value of the banner has to be increased accordingly. So, I am adding breakpoints for small screens so that different sections don't overlap.

assets/scss/_grpc.scss
```diff
@@ -229,17 +234,24 @@ c - Component (Aware of its content/context...)
...
 
   & p {
     padding: 0.5rem;
   }
+  padding-bottom: 0.01rem;
+}
+
+@include media-breakpoint-down(sm){
+  .o-banner{
+    top: 0rem;
```
arjan-bal, line 253: Can you add a comment explaining that this is required because the navbar changes `position` from `fixed` to `absolute`. This behaviour will not be obvious to future readers.

layouts/partials/banner.html
```diff
@@ -6,6 +6,6 @@
       <strong>Register now</strong>
     </a>
     ($50 until Jul 30th)
-  </p>
-</div>
+    </p>
+  </div>
```
arjan-bal, line 10: The indentation changes can be reverted since they were correct earlier.

gnossen (approved):
Nice catch on the mobile aspect, @arjan-bal ! This looks great now. Going ahead and merging.

PR #1332, opened 2024-07-24
Add setup instructions for GA and GTM to README

Gayathri625:
Updated README.md to include setup and verification instructions for Google Analytics and Google Tag Manager.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ $ npm run serve
    ```
     Check GTM setup using Preview mode and verify that Tag Assistant displays "Connected" for your website URL.
    
-Build the site by setting Hugo environment to production to make sure GA and GTM work correctly
+Build the site by setting Hugo environment to production to make sure GA and GTM work.
 ```console
 $ HUGO_ENV=production npm run serve
 ```
--- a/README.md
+++ b/README.md
@@ -60,7 +60,12 @@ $ npm run serve
      gtmID: "GTM-XXXXXXXX" # Replace with your GTM ID
    ```
     Check GTM setup using Preview mode and verify that Tag Assistant displays "Connected" for your website URL.
-    
+   
+Build the site by setting Hugo environment to production to make sure GA and GTM work correctly
+```console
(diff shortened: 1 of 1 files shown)


arjan-bal (changes requested):
Please also mention building with the PRODUCTION variable so that GA works.

gnossen:
Seems like https://github.com/grpc/grpc.io/pull/1332#pullrequestreview-2197470755 still needs to be resolved, so holding off on merging for now.

arjan-bal:
> Please also mention building with the PRODUCTION variable so that GA works.

@Gayathri625  For reference, this seems to be the block that only includes analytics scripts when in production mode: https://github.com/grpc/grpc.io-docsy/blob/ba20af1c9044e3ad8d83ead5153b243c88539038/layouts/partials/head.html#L55-L61

Gayathri625:
> I see that we load GTM scripts in non-prod also in #1313. Can you have a similar prod env check for them also? You can do that in a separate PR.

check for production environment: [#1335](https://github.com/grpc/grpc.io/pull/1335)

README.md
```diff
@@ -39,6 +39,33 @@ To locally serve the site at [localhost:8888][], run the following command:
...
+   - Follow the [GTM setup guide](https://developers.google.com/tag-platform/tag-manager/web) to create your account and container.
+   - Update `config.yaml` with your GTM ID:
+
+   ```yaml
+   params:
+     gtmID: "GTM-XXXXXXXX" # Replace with your GTM ID
+   ```
+    Check GTM setup using Preview mode and verify that Tag Assistant displays "Connected" for your website URL.
+   
+Build the site by setting Hugo environment to production to make sure GA and GTM work correctly
```
arjan-bal, line 64: nit: Can omit "correctly", they will not work at all without it,
>...make sure GA and GTM work.

arjan-bal (approved):
LGTM!

gnossen (approved):

PR #1335, opened 2024-07-26
Add a check for production

Gayathri625:
Load GTM Scripts only when environment is set to production


arjan-bal (approved):
LGTM. Please verify that the GTM scripts are not included when run without prod and included when run in prod mode.

Gayathri625:
> LGTM. Please manually verify that the GTM scripts are not included when run without prod and included when run in prod mode.

Verified, gtm scripts are only loaded when hugo env set to production while building using this
$ HUGO_ENV=production npm run serve

gnossen (approved):

PR #1333, opened 2024-07-24
Update `protoc-gen-go` to 1.34.2 and `protoc-gen-go-grpc` to 1.4.0

nuno-faria:

PR #1338, opened 2024-07-30
grpc-go: update versions

arvindbr8:
Two changes:

1. Since https://github.com/grpc/grpc-go/issues/7249, gRPC-Go only supports the last two versions of Golang. Similar to Golang policy itself
2. Install the latest protoc plugins 

cc: @dfawley

### DIFF
--- a/content/en/docs/languages/go/quickstart.md
+++ b/content/en/docs/languages/go/quickstart.md
@@ -7,7 +7,7 @@ spelling: cSpell:ignore Fatalf GOPATH
 
 ### Prerequisites
 
-- **[Go][]**, any one of the **three latest major** [releases of Go][].
+- **[Go][]**, any one of the **two latest major** [releases of Go][].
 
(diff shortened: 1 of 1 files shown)


ejona86:
https://github.com/grpc/grpc.io/pull/1333 just merged, so this is now in conflict. Do you really not need to specify a version? Will you always make sure to upgrade the grpc version here before doing a new release of the codegen?

arvindbr8:
> #1333 just merged, so this is now in conflict. 

Ah. note to self: check PRs in flight

Do you really not need to specify a version? Will you always make sure to upgrade the grpc version here before doing a new release of the codegen?

for protoc-gen-go-grpc @latest should be the right behavior. 

for protoc-gen-go version - That's right. Our CI checks for a specific version https://github.com/grpc/grpc-go/blob/master/test/tools/go.mod#L8

I'm reverting the conflicting changes and only updating the gRPC-go version policy

ejona86 (approved):

PR #1349, opened 2024-08-22
Update gRPC-core version to 1.66

drfloob:


ejona86 (approved):
1.65 was overrated /s

PR #1354, opened 2024-09-04
feat: move the $ shell line indicator to scss

vordimous:
When tyring to copy cli commands from the docs using a triple click or highlighting the whole line the `$` is included. The same is true using the copy button rendered in the github UI. By making the `$` line indicator for shell code part of the CSS it removes the accidental copy of `$` and makes following guides more user friendly.

before line highlight after triple click:
image

with the CSS implemented `$`:
image


sreenithi:
Hi @vordimous, thank you for these changes. Looks like there is a merge conflict in `quickstart.md`. Please resolve it too

vordimous:
@sreenithi I have updated the code to handle the conflict and found the remaining `$` in code blocks. I also noticed that powershell commands were added so I included a change to automatically include the `>` for powershell code blocks

sreenithi (approved):
LGTM, @shivaspeaks just adding you for an additional review as there are quite a lot of files changed

shivaspeaks (commented):
Hmmm. The description says suggestion is to make the $ shell prompt a CSS-only visual element rather than being part of the actual text in the HTML content. This ensures the $ is visible in the rendered documentation but excluded from copied text.

But I don't see `$` getting rendered with the commit: [5fed470197fe68e6cc2248f4862fcf5a4ed27b1f](https://github.com/grpc/grpc.io/blob/5fed470197fe68e6cc2248f4862fcf5a4ed27b1f/content/en/blog/bazel-rules-protobuf.md)

vordimous:
also note that the css is only looking for \`\`\`sh and \`\`\`powershell codeblocks. This means any codeblock you would want to keep as is would use \`\`\`bash

shivaspeaks (approved):

PR #1355, opened 2024-09-09
Update auth.md

LikDan:


ejona86:
@arjan-bal, can you review since this is for Go?

content/en/docs/guides/auth.md
```diff
@@ -213,8 +213,8 @@ and authorization in various languages.
 | Python   | [Python Example]                          | [Python Documentation] |
 
 
-[Go Example]: https://github.com/grpc/grpc-go/tree/master/examples/features/encryption
-[Go Documentation]: https://github.com/grpc/grpc-go/tree/master/examples/features/encryption#encryption
+[Go Example]: https://github.com/grpc/grpc-go/tree/master/examples/features/authentication
+[Go Documentation]: https://github.com/grpc/grpc-go/tree/master/examples/features/authentication#encryption
```
arjan-bal, line 217: Since there is no `#encryption` section on the page, this link seems broken.

arjan-bal, line 217: This table is for language specific TLS examples. So the Go links seems to correctly point to the encryption directory. The table just below this is for OAuth examples, in which the Go row points to the links you've added in this PR.
https://github.com/grpc/grpc.io/blob/f4a96fb104532683f6fcf68408268550ef0c1ba6/content/en/docs/guides/auth.md?plain=1#L224-L238

arjan-bal (commented):
There are separate tables for OAuth and TLS examples, and links for Go seem to be correct in both. I don't think any changes are required here.

Reallyfe (approved):

Reallyfe (approved):

PR #1356, opened 2024-09-09
update description of retry backoff behavior

markdroth:
Updates in accordance with https://github.com/grpc/proposal/pull/452.

@ejona86

### DIFF
--- a/content/en/docs/guides/retry.md
+++ b/content/en/docs/guides/retry.md
@@ -63,7 +63,7 @@ Retry is configurable via [gRPC Service Config], at a per-method granularity.
 }
 ```
 
-The actual initial backoff delay is a random time period within `initialBackoff` value configured above (between 0-100ms).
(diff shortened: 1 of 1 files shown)


content/en/docs/guides/retry.md
```diff
@@ -63,7 +63,7 @@ Retry is configurable via [gRPC Service Config], at a per-method granularity.
 }
 ```
 
-The actual initial backoff delay is a random time period within `initialBackoff` value configured above (between 0-100ms).
+Jitter of plus or minus 0.2 is applied to the backoff delay to avoid hammering servers at the same time from a large number of clients.  In the example configuration above `initialBackoff` is set to 100ms, so the actual backoff delay after the first attempt will be for a random time period within the range `[80ms, 120ms]`.
```
dfawley, line 66: Suggestion: `20%` instead of `0.2` would be less possible to confuse ("`0.2` what? seconds?").

dfawley (approved):

content/en/docs/guides/retry.md
markdroth: Good idea, done.

ejona86 (approved):

PR #1357, opened 2024-09-09
dualstack user guide

markdroth:
@temawi

### DIFF
--- a/content/en/docs/guides/dualstack.md
+++ b/content/en/docs/guides/dualstack.md
@@ -1,0 +1,55 @@
+---
+title: "Dualstack Backends"
+description: >-
+     Explains the dualstack backend feature in gRPC.
+---
+
+### Overview
+
+In cases where an endpoint (i.e., a gRPC server) has both an IPv4 and
+IPv6 address, it is useful for clients to know both of those addresses,
+so that they can reach the endpoint on either of them.  Because only
+one of the two protocols may actually be routable between a given
+client and endpoint, clients need to try both addresses to see which
+one will work.  To minimize connection attempt latency, it is also
+desirable for clients to attempt to connect to multiple addresses in
+parallel, which is done using the "Happy Eyeballs" algorithm described in
+[RFC-8305](https://www.rfc-editor.org/rfc/rfc8305).
+
+In gRPC, it is particularly important for the [Load Balancing
+Policy](https://grpc.io/docs/guides/custom-load-balancing) to understand
+which addresses are associated with the same endpoint as opposed to
+being different endpoints, so that the client does not send twice as
+much load to a endpoint with two addresses.  To that end, the [Name
+Resolver](https://grpc.io/docs/guides/custom-name-resolution) can return
+multiple addresses per endpoint (where an endpoint is a single endpoint),
+and that information is communicated to the Load Balancing Policy.
+
+Note that the only name resolver implementation that gRPC provides is a
+DNS resolver, which does not support dualstack backends.  This is
+because DNS inherently returns a flat list of addresses, with no way to
+indicate whether or not multiple addresses actually point to the same
+endpoint.
+
+There are two ways to use dualstack endpoints in gRPC:
+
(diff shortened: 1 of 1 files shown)


markdroth:
@ejona86 @dfawley I'd welcome suggestions for other details to cover here.  There are certainly a lot of underlying details, but I'm not sure how many of them are actually important for someone trying to use this feature.

temawi:
This looks ok to me, but I'll hold off from merging for now to give @ejona86 and @dfawley a chance to take a look.

temawi (approved):

content/en/docs/guides/dualstack.md
```diff
@@ -0,0 +1,55 @@
...
+Policy](https://grpc.io/docs/guides/custom-load-balancing) to understand
+which addresses are associated with the same endpoint as opposed to
+being different endpoints, so that the client does not send twice as
+much load to a endpoint with two addresses.  To that end, the [Name
+Resolver](https://grpc.io/docs/guides/custom-name-resolution) can return
+multiple addresses per endpoint (where an endpoint is a single endpoint),
+and that information is communicated to the Load Balancing Policy.
+
+Note that the only name resolver implementation that gRPC provides is a
+DNS resolver, which does not support dualstack backends.  This is
```
ejona86, line 29: It won't be clear to users what "does not support" means here. You can use dualstack backends with pick_first. Maybe the earlier paragraph about load balancing policies needs to call out that nothing special needs to happen for the default policy, pick_first. But if using a different policy it probably needs NR support? And then you can use round_robin as an example when explaining the twice-the-load problem.

content/en/docs/guides/dualstack.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Overview
+
+In cases where an endpoint (i.e., a gRPC server) has both an IPv4 and
+IPv6 address, it is useful for clients to know both of those addresses,
+so that they can reach the endpoint on either of them.  Because only
+one of the two protocols may actually be routable between a given
+client and endpoint, clients need to try both addresses to see which
+one will work.  To minimize connection attempt latency, it is also
+desirable for clients to attempt to connect to multiple addresses in
```
ejona86, line 15: "it is also desirable" doesn't say what is actually done or who is responsible for it. "To minimize connection attempt latency, gRPC clients will attempt..."?

content/en/docs/guides/dualstack.md
```diff
@@ -0,0 +1,55 @@
...
+endpoint.
+
+There are two ways to use dualstack endpoints in gRPC:
+
+1. Write your own custom name resolver that returns multiple
+   addresses per endpoint.  For more information, see [Custom Name
+   Resolution](https://grpc.io/docs/guides/custom-name-resolution).
+2. Use an xDS control plane that can communicate multiple addresses
+   per endpoint to the client.  For more information, consult the
+   documentation for your xDS control plane.
```
dfawley, line 41: I was _going to say_ this should link to some grpc xds documentation for how to set up your client to talk to your control plane, but...I can't find anything appropriately user-facing.  Should we link to https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md for now?

Should we create another guide that covers how to set up a bootstrap file and create an xds-enabled grpc client (and server, possibly separately)?

content/en/docs/guides/dualstack.md
```diff
@@ -0,0 +1,55 @@
...
+For more information on gRPC dualstack backend support, see [gRFC
+A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md).
+
+### Custom Resolver Dualstack Examples
+
+| Language | Example          | Notes                                  |
+|----------|------------------|----------------------------------------|
+| C++      |                  | C++ resolver API not yet public.       |
+| Go       |                  | Go dualstack support not yet complete. |
+| Java     | [Java example]   |                                        |
```
dfawley, line 52: Maybe indicate that both Java and Go have only experimental support for custom LB policies?

PR #1362, opened 2024-09-19
Update logo.svg

sahilk:
Cleaned up the diamond shapes. Too many elements in the older version. Reduced file size as well as a result.


sreenithi:
hi @sahilk, can you elaborate on why this update is needed?

sahilk:
Not needed. Just a clean up. I was looking to use the logo as part of some documentation and realised there were too many layers and clips in the avg making it unnecessarily large and messy. You can choose to reject it.

ejona86:
The preview makes it look like the new file deletes a transparency gradient.

sahilk:
@ejona86 Yeah, GH's rendering is showing the dithering for some reason. The gRPC website doesn't show this detail however and I only realized there were all these layers when I opened the SVG in Figma and tried selecting a layer. IMO, it doesn't need all the layers.

ejona86:
I agree that I don't see transparency on grpc.io, but I do see a gradient. I do agree this modified one is cleaner in the file, but it is also a different image. 19 KiB is a small enough image these days and that is much smaller than the png we use, that it's not worth optimizing if it changes the resulting images.

Thank you all the same!

PR #1397, opened 2025-01-10
Add Windows as a supported platform for PHP

SRWieZ:
PHP Windows DLLs are available.

- https://pecl.php.net/package/grpc
- https://github.com/grpc/grpc/tree/master/src/php


stanley-cheung (commented):
Thanks for the patch. The DLLs were automatically built by the PECL repository but our project currently does not have the resource to support Windows as a platform.

SRWieZ:
Thanks for the explanation.

From a user perspective, I don't see the difference. The PECL-generated DLL works too, right?

If not appropriate you can close this PR.

PR #1400, opened 2025-01-16
docs: add roadrunner as an option for implementing a gRPC PHP server

bshaffer:
fixes https://github.com/grpc/grpc/issues/38481


shivaspeaks:
@stanley-cheung can you review this?

PR #1402, opened 2025-01-17
Update Java version to 1.69.1

ejona86:


ejona86:
#1405 has already upgraded to even a later version.

PR #1440, opened 2025-06-18
Update introduction.md

Logofile:
Updates recommendation for using proto3 to use Editions instead.


ejona86:
Given that not all languages support editions yet, it seems we don't actually recommend it yet.

ejona86:
The ones I'm aware of:
https://github.com/grpc/grpc-java/issues/11526
https://github.com/grpc/grpc-kotlin/issues/625

PR #1446, opened 2025-07-07
Remove Meet page

gnossen:


ejona86 (approved):

ejona86:
It looks like we won't have anyone reviewing the form submissions for a while. Thus removing the page. We can always add it back if someone will coordinate it.

PR #1447, opened 2025-07-10
Add Swift to languages

glbrntt:
Swift has the hello world and route guide tutorials host externally so this change also introduce an external docs to the config and uses that in the language home pages if present.


shivaspeaks:
CC: @thisisnotapril

glbrntt:
@thisisnotapril, could you take a look at this please? 🙏

sreenithi (approved):

PR #1448, opened 2025-07-11
[OTel Metrics] Format page

yashykt:


ejona86 (changes requested):
Remove the mdformat goo. That seems to be a feature of your custom formatter, which isn't relevant to any other person modifying this file.

I don't know why the formatter is putting three spaces after bullets. I can't say I've seen that style before. (But whatever.)

content/en/docs/guides/opentelemetry-metrics.md
```diff
@@ -5,149 +5,184 @@ description: OpenTelemetry Metrics available in gRPC
...
+*   [A66: OpenTelemetry Metrics]
+*   [A78: gRPC OTel Metrics for WRR, Pick First, and XdsClient]
+*   [A79: Non-per-call Metrics Architecture]
 
+[sunsetted]: https://opentelemetry.io/blog/2023/sunsetting-opencensus/
+[MeterProvider]: https://opentelemetry.io/docs/specs/otel/metrics/api/#meterprovider
+[OpenTelemetry API]: https://opentelemetry.io/docs/specs/otel/overview/#api
+[Meter]: https://opentelemetry.io/docs/specs/otel/metrics/api/#get-a-meter
+[OpenTelemetry SDK]: https://opentelemetry.io/docs/specs/otel/overview/#sdk
+[C++ Example]: https://github.com/grpc/grpc/tree/master/examples/cpp/otel
```
ejona86, line 175: These examples are already defined below.

ejona86 (approved):
FWIW, things like "sunsetted" we tend to keep in the same section it is used (I do it for most links used only once). But at the end is technically acceptable.

One thing to fix

yashykt:
> FWIW, things like "sunsetted" we tend to keep in the same section it is used (I do it for most links used only once). But at the end is technically acceptable.

I'm fine either way. I don't mind someone reverting that to how it was previously either.

PR #1457, opened 2025-07-31
Update banner 

cjqzhao:
Update banner to get rid of early bird pricing


sreenithi:
Can you also include information about gRPConf India CFP and early bird pricing open

sreenithi (approved):

PR #1473, opened 2025-10-24
Update async.md

abuhelos:
In multithreaded implementations, status_ needs to be updated before instructions that lead to CallData being put back into the completion queue. Otherwise, a race condition can occur where thread_1 picks up a Call Data instance that is intended to execute the current state (i.e. FINISH) but instead executes the previous state (i.e. PROCESS) because the status was not updated yet by thread_2.

### DIFF
--- a/content/en/docs/languages/cpp/async.md
+++ b/content/en/docs/languages/cpp/async.md
@@ -136,6 +136,11 @@ public:
 
   void Proceed() {
     if (status_ == CREATE) {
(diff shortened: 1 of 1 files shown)


rajatnaik0405:
@ctiller Can you please review this ?

abuhelos:
@rajatnaik0405 @ctiller Just wanted to follow up on this. I have this noted in my own project docs but would also like to confirm this to ensure my understanding is accurate here.

PR #1506, opened 2026-05-01
update the grpc java version to 1.81.0 in params.grpc_vers.java

AgraVator:


ejona86 (approved):

ejona86:
@AgraVator, your base commit was quite old so there were conflicts. Next time make sure to begin with a semi-recent commit.

